PR #132, opened 2015-02-02
Fix Index For Read

einentwickler:
Assign indexBaseForLine to dataOffsetStart when (indexLineEntry==0)
instead of (inLine==0)


lburgazzoli:
Hi, do you have a test case for this issue ?

PR #134, opened 2015-02-10
Fix indexForRead with IndexedChronicle test case for large file

einentwickler:
Fix indexForRead with IndexedChronicle test case for large file


einentwickler:
hi, i have include both the fix and the test case in the same commit.
So please refer to this PR, and the old PR is removed.

lburgazzoli:
Thank you for the PR

PR #191, opened 2015-09-15
Working directory support implemented for Vanilla ExcerptTailer. Fetc…

pisfly:


lburgazzoli:
Can you please add a test case for this ? Is is a trivial change but we'd like to have a test case for any new feature so we can detect if we inadvertently break it.

pisfly:
Hi, 
I added test case for the function.

lburgazzoli:
Hi. one more request: can you please update this PR with the latest source ?

pisfly:
Hi, updated it.

lburgazzoli:
Thx

pisfly:
Hello,

I see these changes now in chronicle-queue-4.0.1-alpha. Where can I get the snapshot and also stable to use it?

lburgazzoli:
Queue v4 ha a complete different implementation, you changes are in 3.6.0-SNAPSHOT you can get from Sonatype OSS

Stable will be released as soon as possible

RobAustin:
I suggest you stick with chronicle-queue v3 for now, as version 4 is in alpha, it should be out of alpha towards the end of the year.

pisfly:
Thanks Rob. I think I can continue with 3.6.0-SNAPSHOT which includes changes from this PR and also v3 as you mention?

RobAustin:
Are you looking for  3.6.0-SNAPSHOT to be release (asap), if you are I'll, see what we can do.

Rob

pisfly:
We have 2 or 3 weeks before our release so It would be nice to have it.

PR #270, opened 2016-07-22
Added failing test for BackwardsRollIssue.

scottkidder:
Test fails most of the time, but still not always when running 2M messages through it.

A tailer moves back to the first index ever read, sometimes both tailers at once.


scottkidder:
I have traced this to the method SingleChronicleQueueExceprts.StoreTailer.next().

When this.store == null, the queue immediately get's the firstIndex and moves to it.

For a reason which I cannot find, the store is null sometimes and this when the roll backwards occurs.

RobAustin:
merged tests

PR #271, opened 2016-07-27
Backwards roll failing tests

scottkidder:
Both tests failing

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
             <dependency>
                 <groupId>net.openhft</groupId>
                 <artifactId>chronicle-bom</artifactId>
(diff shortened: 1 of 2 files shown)


scottkidder:
Second commit is a working fix for when the store becomes null, rather than going to firstIndex(), attemptToAdvanceCycle(). All tests are passing but handling a BACKWARD tailer should be cleaner.

peter-lawrey:
This test has been added. Can you re-issue this PR if there is anything missing?

peter-lawrey:
I don't think this has been pulling manually after all.

peter-lawrey:
Would you be able to update and re-issue this PR?

PR #282, opened 2016-08-06
Extended the WriteReadTextTest

apete:
Merged my test code with what Peter had already written.


apete:
This pull request contains test cases that reproduce both "not getting back the same string" and "crashing the JVM".

(Not a GIT expert - don't know how to get rid of those "merge" commits.)

PR #288, opened 2016-08-11
Use exceptionHandlers in toEndTest.

scottkidder:
Record DEBUG level exceptions because MappedFile.close() throws IllegalStateException - Released at the DEBUG level.


scottkidder:
* Some of the things being logged at DEBUG level are not even Exceptions, moveToIndex, Allocation of chunk, etc.
* Jvm.recordExceptions(false) removes the DEBUG logger and replaces it with NullExceptionHandler.NOTHING, which causes the messages to not be recorded or logged to console.

PR #294, opened 2016-09-15
SCQS.writePosition padding

scottkidder:
Reads of SCQS.writePosition were causing 'WARN: add padding for concurrent appenders...' message. After padding the warning is gone.


scottkidder:
Ref: http://psy-lob-saw.blogspot.com/2013/02/alignment-concurrency-and-torture-x86.html

scottkidder:
After testing, I've found that padding in place caused the WARN messages again. Need to investigate further.

PR #303, opened 2016-10-19
Test shows problems with invocation of method without parameters.

fast-runner:
Test that shows problems with invocation of method without parameters.

I hope this help in issue fixing :)


peter-lawrey:
This has been fixed in the current SNAPSHOT.

Issue #306, opened 2016-11-07
Possible file descriptor leak


amichair:
We've had some trouble with cycled files that are not being released. In a heap dump I see a ThreadLocal with SCQIndexing$LongArrayValuesHolder -> BinaryLongArrayReference -> MappedBytes -> MappedFile -> RandomAccessFile (and channel) holding on to a file that was cycled out and (I think) should have been released but isn't. The MappedFile refCount is at 3, although there are 9 references to the MappedFile (including 7 MappedBytes instances).

When should a file be ultimately released? Will the WeakReference in #305 and elsewhere dispose of the object reference but leave the file open? Does a reference count need to be released first? It's also possible that there's a ref count leak regardless of the ThreadLocal, or that I'm simply doing something wrong or misunderstanding the file descriptor management here...

As @peter-lawrey mentioned in #305, at the very least there should be tests for this.

peter-lawrey:
There should be tests.

The reference count is intended to provide deterministic clean up. If the
objects are discarded first they will be cleaned up by the GC, at some
point.

We are very busy with paying clients until mid December so I am not sure
when will be able to look at this in detail. I am sorry we don't have spare
resources at the moment.

Regards, Peter.

On 7 Nov 2016 18:50, "amichair" <notifications@github.com> wrote:

We've had some trouble with cycled files that are not being released. In a
heap dump I see a ThreadLocal with SCQIndexing$LongArrayValuesHolder ->
BinaryLongArrayReference -> MappedBytes -> MappedFile -> RandomAccessFile
(and channel) holding on to a file that was cycled out and (I think) should
have been released but isn't. The MappedFile refCount is at 3, although
there are 9 references to the MappedFile (including 7 MappedBytes
instances).

When should a file be ultimately released? Will the WeakReference in #305
<https://github.com/OpenHFT/Chronicle-Queue/issues/305> and elsewhere
dispose of the object reference but leave the file open? Does a reference
count need to be released first? It's also possible that there's a ref
count leak regardless of the ThreadLocal, or that I'm simply doing
something wrong or misunderstanding the file descriptor management here...

As @peter-lawrey <https://github.com/peter-lawrey> mentioned in #305
<https://github.com/OpenHFT/Chronicle-Queue/issues/305>, at the very least
there should be tests for this.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
[...]

PR #314, opened 2016-11-29
fixed store reference leaks

amichair:
This is another step in solving #306 : some store usages never call release, thus preventing the reference count from reaching zero and the resources from being freed.


peter-lawrey:
Thank you for the fix.

PR #318, opened 2017-01-04
Failing read only test: tailer can't get toEnd()

andrm:
If I switch a queue to read-only, tailers can't go to the end anymore. I have difficulties understanding the read-only mode.


peter-lawrey:
Read only mode needs more tests. Thank you.

PR #320, opened 2017-01-09
Trailer -> Tailer

andrm:
Trailer -> Tailer: Chronicle Queue uses the term "Tailer", not "Trailer"


peter-lawrey:
Thank you. FIX uses the term Trailer. We have been updating the docs of both.

PR #323, opened 2017-01-10
Tailer can't read forward again after having hit start of queue readi…

andrm:
…ng backwards.

(Tested against 4.2.21, since 22 and 23-SNAPSHOT don't work right now)

I was wondering if this is supposed to be working. The failure is in line 125, reading forward is not possible anymore.

 I think I had the same problem when I did a "toStart()" and tried to start reading.


peter-lawrey:
This should be fixed  now.

PR #331, opened 2017-01-24
Failing Test with successive appenders.

andrm:
I was expecting that the second appender would just add entries, but it rewrote them. Anything I am missing?


peter-lawrey:
Each thread has only one appender as there is only one end of file. This does mean you can't track the last index appended without taking a copy.

peter-lawrey:
Thank you for the test.
If you do an update, I have modified the test and added some comments.
Could you review them?

andrm:
I'm not sure what you mean: A copy of the long (lastIndexAppended) or the appender?

peter-lawrey:
Sorry, I hadn't read the test properly. You have two actual appenders in this case as you have two queues.

andrm:
Sorry, yes, looks good.

andrm:
I do have another problem with 4.5.21 I'm trying to reproduce, but wasn't able to do yet. I will post a test once I can. Thanks!

peter-lawrey:
Thank you. The tests are very helpful in turning these problems around.
ᐧ

andrm:
Thanks for the fast responses.

PR #332, opened 2017-01-25
Failing test with runaway index.

andrm:
Double close on DocumentContext moves the index by 2, not 1. Result: Appenders and Tailers disagree on the index position.

Please fix this. It cost me so much time to isolate.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/OvertakeTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/OvertakeTest.java
@@ -1,0 +1,100 @@
+package net.openhft.chronicle.queue;
+
+import net.openhft.chronicle.core.OS;
+import net.openhft.chronicle.core.io.IOTools;
+import net.openhft.chronicle.queue.impl.single.SingleChronicleQueue;
+import net.openhft.chronicle.queue.impl.single.SingleChronicleQueueBuilder;
+import net.openhft.chronicle.wire.DocumentContext;
+import net.openhft.chronicle.wire.Wire;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Index runs away on double close - AM
+ */
+public class OvertakeTest {
+
+    private String path;
+
+    private long a_index;
+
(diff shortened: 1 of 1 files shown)


peter-lawrey:
This has been fixed so duplicate calls to close() are ignored.

peter-lawrey:
Thank you.

andrm:
You're welcome. 
- Any release soon?
- Is there an irc or slack channel with you guys? Sometimes I have questions wrt chronicle how it's supposed to be working and the google group is a bit too heavy for me.

peter-lawrey:
We have direct email and phone for paid support.
For free support we prefer that every one see the answers so all can
benefit from them.
ᐧ

andrm:
I understand. I tried to contact Chronicle Software for paid support but I didn't hear back.

BTW: Double close had a unit test in November in the google group under "Chronicle Queue v4 writes failing due to seq mismatch" Why wasn't it integrated?

peter-lawrey:
Can you try sales@chronicle.software? I have checked the email for this
account and I couldn't find your name.
ᐧ

peter-lawrey:
I am not sure issue you mean but a bug was reported and fixed.
The error message s/he reported was the result of a diagnostic trying to
track down another related issue.

In the end it turned out that CAS operations which split CPU cache are not
atomic and fail silently on x64. (On ARM they fail with a signal)
To avoid this we add padding to ensure the 4 byte header is always within a
cache line.
ᐧ

andrm:
From the post:
Kyle Kavanagh
> See attached.  In creating this pared-down test, I actually think I discovered my error.  I was using a tailer DocumentContext in a try-with-resources block, but was closing the context manually inside the block.  **This would mean that close() would be called twice on the context.**  What made this tricky to track down is that the error/stacktrace only appears on the writer thread, which lead me to think that it was an issue with the writer side of the queue.  I realized that it was the tailer when I wasn't able to recreate the issue at first and observed that it only seems to be a problem when another thread is actively tailing the queue.   **A fix for this would just be to make close() a noop if already closed.**

Rob Austin:

> Thanks. - We will implement your suggestion. 
> 
> Rob


I guess it fell through the cracks.

peter-lawrey:
Hmmm. I think a fix was applied but as you found it didn't cover this
properly.

PR #337, opened 2017-02-04
Second Question

Catherine-Lawrey:
My version of the page wasn't updated with the changes I made to the actual website page. So those changes have been made here. 
I've also added another question about using Chronicle Queue instead of RMI


peter-lawrey:
I have made some changes, such as update for Single Chronicle.
Thank you.

PR #338, opened 2017-02-06
2 questions added to the end

Catherine-Lawrey:
Questions from http://stackoverflow.com/questions/38415008/java-chronicle-throwing-illegalstateexecption-at-high-number-of-messages and http://stackoverflow.com/questions/41429878/chronicle-queue-events-listener


peter-lawrey:
Thank you

PR #355, opened 2017-06-25
Fix resource leaks caused by WireStore refCounting errors

sergebg:
To find leaks I run some simple scenario and then check open file descriptors using linux "lsof" command. At first I suspected that leaks were caused by the direct call "store.release()". But the actual problem is "reserve" after "acquire".


scottkidder:
Can someone review/merge this please? Some of my nagging issues appear to be resolved from my initial testing.

PR #358, opened 2017-06-29
Test for too many files problem

sergebg:
That is one more form of resource leaks. Closers are stored in WeakHashMap, and so from time to time gc() just eats them. In my application I need constant set of trailers and appenders, so I just replaced WeakHashMap with HashMap. In general case it makes sense to apply some finalisation approach, e.g.  phantom references.


MartyIX:
This is my output for the test: https://pastebin.com/0UPFycYV

PR #398, opened 2017-11-06
Queuelock implementation for replication

dpisklov:


src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public TSQueueLock(File queueDirectoryPath, Supplier<Pauser> pauser) {
+        final File storeFilePath;
+        if ("".equals(queueDirectoryPath.getPath())) {
+            storeFilePath = new File(QUEUE_LOCK_FILE);
+        } else {
+            storeFilePath = new File(queueDirectoryPath, QUEUE_LOCK_FILE);
+            queueDirectoryPath.mkdirs();
+        }
+        TableStore tableStore = SingleTableBuilder.binary(storeFilePath).build();
```
epickrram: I think this is prone to concurrent threads creating multiple instances of the LOCK_KEY reference in the table.

I had to use file locking to get around this for the directory listing: https://github.com/OpenHFT/Chronicle-Queue/blob/master/src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java#L51

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
```diff
@@ -0,0 +1,128 @@
...
+     * Stores current PID to table store, and any other process trying to acquire lock will wait for
+     * <code>chronicle.queue.lock.timeoutMS</code> millis (default is 30000) for the lock to be released, and if it is not
+     * after timeout, throws {@link IllegalStateException}. Also the locking thread ID is stored in threadLocal field, so
+     * that only locking thread is allowed to {@link net.openhft.chronicle.queue.ChronicleQueue#acquireAppender} while the queue is locked.
+     */
+    @Override
+    public void acquireLock() {
+        long tid = Thread.currentThread().getId();
+        try {
+            while (!lock.compareAndSwapValue(UNLOCKED, PID)) {
```
epickrram: Sorry, I must be being daft - how does this exit if the current process fails to lock the queue? I can't fathom where the TimeoutException is thrown from.

dpisklov: pauser.pause throws TimeoutException

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
dpisklov: Hmm I didn't check where exactly you did it, I assumed it was fixed for table store as a whole - why don't we move this to main TableStore implementation to avoid duplication of code?

epickrram: The pattern so far is to use a lock around initial calls to acquireKey, then just use the LongValue references. If the file-lock is moved to acquireKey() in table store, then it will be invoked each time. 
I think the change to move it there would be ok, but it may be worth overloading the method and calling it acquireKeyWithExclusiveLock or something similar.Hm. The pattern so far is to use a lock around initial calls to acquireKey, then just use the LongValue references. If the file-lock is moved to acquireKey() in table store, then it will be invoked each time. 
I think the change to move it there would be ok, but it may be worth overloading the method and calling it acquireKeyWithExclusiveLock or something similar.

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
epickrram: Oh, the Pauser is stateful - that's delightful.

Probably not safe for access from multiple threads then...

dpisklov: Just like 99% of queue code :) Not sure we really want to handle multithreading here - essentially there shouldn't be anyone trying to acquire lock other then our own replication Source.

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
dpisklov: Changed to use locking

src/main/java/net/openhft/chronicle/queue/impl/single/NoopQueueLock.java
```diff
@@ -0,0 +1,35 @@
...
+    public void checkLock() {
+    }
+
+    @Override
+    public void acquireLock() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void unlock() {
```
RobAustin: question should this throw UnsupportedOperationException or just not do anything ?

PR #441, opened 2018-02-26
Update ChronicleReaderMain.java

danielshaya:
At the moment a custom plugin must be set - this allows the Reader to function even without the -p option being set.


michaelszymczak:
+1 As this mandatory flag actually prevented from using the Reader in the existing default way (printing messages to the stdout).

PR #468, opened 2018-05-07
Disk space logging executor service is now configurable

### LINKED ISSUE #467: Paramaterize disk space checker
Hi,

Is there interest in parameterizing the `ExecutorService` that the disk space checker runs on? (https://github.com/OpenHFT/Chronicle-Queue/blob/master/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java#L63)

We're using CQ in a resource constrained environment and already have an `ExecutorService` that we use for blocking operations. It would be great if we could reuse this and save the 1MB thread call stack.

I'd be happy to explore implementing this and submitting a PR; I just wanted to gauge interest first before I go down that route.

Thanks!

longshorej:
This makes the `ExecutorService` that the disk space checker runs on configurable. The built-in thread pool is now lazily initialized so that it is only created if it is needed.

Manually verified via taking a thread dump and looking at the running threads. When providing my own executor service, the default thread "disk-space-checker" isn't created.

Fixes #467


dpisklov:
@longshorej 
Hi, 
I think there's some inconsistency in this code. Builder is used for the stuff that belongs to instance, not to the static field. What about leaving default static as it was but make it non-final and add a static method to set it? That way you would be able to call it before you use the queue to set your own ES.

longshorej:
Thanks for the feedback @dpisklov. I was thinking that it could be configurable per queue, but I'd be happy to make it global instead. I'll update this PR shortly.

dpisklov:
@longshorej @cmunger this change has been overwritten recently. Now to achieve the same you need to use `net.openhft.chronicle.threads.Threads#executorFactory()` method (the code is in OpenHFT/Chronicle-Threads project). It's fairly self-explanatory but if you have any questions do let me know.
Additionally this change affects Ref cleaner thread.

longshorej:
Thanks for the heads up @dpisklov

PR #574, opened 2019-01-31
Polish

JanStureNielsen:
Switch from deprecated `buffered(boolean buffered)` to `writeBufferMode(BufferMode mode)`.


RobAustin:
Thanks for this pull request but we will have to review the impact of this change on chronicle-queue-enterprise before we can merge it in.

PR #586, opened 2019-02-04
Polish: suppress restricted access warnings

JanStureNielsen:


dpisklov:
Don't see these warnings with Java 8, and that's our main supported version.

PR #588, opened 2019-02-04
Polish: remove deprecated method call

JanStureNielsen:
Removing the deprecated call does not change the test outcome.


dpisklov:
This breaks the test

JanStureNielsen:
Ahh -- dang -- sorry about that. I'll track down why this didn't get flagged on my build+test cycle...

dpisklov:
Because (I guess) the test still indicated as passed - despite AssertionError being printed. That's due to the way MethodWriter works...

Issue #620, opened 2019-10-08
java.time.DateTimeException when using a custom RollCycle with a pattern without time


nanomad:
We are experiencing this exception when using a CQ with a custom, weekly, RollCycle configured with the yyyyDDD format.

`java.time.DateTimeException: Unable to obtain Instant from TemporalAccessor: {},ISO,UTC resolved to 1970-01-01 of type java.time.format.Parsed

	at java.time.Instant.from(Instant.java:378)
	at net.openhft.chronicle.queue.impl.RollingResourcesCache.toLong(RollingResourcesCache.java:132)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.cycleTree(SingleChronicleQueue.java:889)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.nextCycle(SingleChronicleQueue.java:916)
	at net.openhft.chronicle.queue.impl.WireStorePool.nextCycle(WireStorePool.java:106)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue.nextCycle(SingleChronicleQueue.java:483)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueExcerpts$StoreTailer.nextIndexWithNextAvailableCycle(SingleChronicleQueueExcerpts.java:1326)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueExcerpts$StoreTailer.endOfCycle(SingleChronicleQueueExcerpts.java:1149)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueExcerpts$StoreTailer.next0(SingleChronicleQueueExcerpts.java:1124)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueExcerpts$StoreTailer.readingDocument(SingleChronicleQueueExcerpts.java:1050)
	at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueExcerpts$StoreTailer.readingDocument(SingleChronicleQueueExcerpts.java:980)
	at net.openhft.chronicle.wire.MarshallableIn.readText(MarshallableIn.java:96)`

This derives from net.openhft.chronicle.queue.impl.RollingResourcesCache#toLong not properly handling this case. 
We have fixed this internally and will submit a tenative PR for it

dpisklov:
@nanomad thanks for reporting and investigation, much appreciated!

[...]

PR #621, opened 2019-10-08
Fix RollingResourcesCache#toLong to support length > ONE_DAY_IN_MILLI…

nanomad:
Closes https://github.com/OpenHFT/Chronicle-Queue/issues/620


dpisklov:
Merged - thanks

PR #654, opened 2020-04-13
Make help printed by default (no args) for CLI tools

dkhamitov:
Very minor but maybe it would be a bit nicer to have something more descriptive than just usage printed when no arguments are provided. It seems to be a de facto standard to do so if a tool doesn't have lots of options and huge help info. E.g., off the top of my head, CLI such as java, groovy, sdkman do this. Code is based on http://commons.apache.org/proper/commons-cli/usage.html

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/ChronicleHistoryReaderMain.java
+++ b/src/main/java/net/openhft/chronicle/queue/ChronicleHistoryReaderMain.java
@@ -79,22 +79,30 @@ public class ChronicleHistoryReaderMain {
             commandLine = parser.parse(options, args);
 
             if (commandLine.hasOption('h')) {
-                new HelpFormatter().printHelp(ChronicleHistoryReaderMain.class.getSimpleName(), options);
-                System.exit(0);
+                printHelpAndExit(options, 0);
             }
         } catch (ParseException e) {
-            printUsageAndExit(options);
+            printHelpAndExit(options, 1);
         }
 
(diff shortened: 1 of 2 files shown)


dkhamitov:
Before:
```
java -cp ... ChronicleReaderMain

Please specify the directory with -d

usage: ChronicleReaderMain [-d <directory>] [-e <exclude-regex>] [-f] [-h] [-i <include-regex>] [-l] [-m <max-history>] [-n <from-index>] [-r] [-s] [-w <wire-type>]
```
After:
```
java -cp ... ChronicleReaderMain

Please specify the directory with -d

usage: ChronicleReaderMain [-d <directory>] [-e <exclude-regex>] [-f] [-h] [-i <include-regex>] [-l] [-m <max-history>] [-n <from-index>] [-r] [-s] [-w <wire-type>]
 -d <directory>       Directory containing chronicle queue files
 -e <exclude-regex>   Do not display records containing this regular expression
 -f                   Tail behaviour - wait for new records to arrive
 -h                   Print this help and exit
 -i <include-regex>   Display records containing this regular expression
 -l                   Squash each output message into a single line
 -m <max-history>     Show this many records from the end of the data set
 -n <from-index>      Start reading from this index (e.g. 0x123ABE)
 -r                   Use when reading from a queue generated using a MethodWriter
 -s                   Display index
 -w <wire-type>       Control output i.e. JSON
```

PR #758, opened 2020-11-21
javadoc. Ensure sink replication also locks

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -530,7 +530,6 @@ public void writeBytes(@NotNull final BytesStore bytes) throws UnrecoverableTime
      */
     public void writeBytes(final long index, @NotNull final BytesStore bytes) {
         throwExceptionIfClosed();
-        checkAppendLock();
         writeLock.lock();
```
JerryShea, line 533: removed as writeBytesInternal calls. Or did you want to lock in case a non-sink calls this?

JerryShea: actually I will put back

JerryShea: Ignore this comment

PR #795, opened 2021-01-11
Undeprecate ChronicleReaderMain. Fix #792, Fix #794

### LINKED ISSUE #792: Remove code duplication between ChronicleHistoryReader and InternalChronicleHistoryReader
It maybe that - it is difficult to dedupe chronicle history reader because it was designed to be extensible. One option could be to leave as it was and remove the duplicated internal version. What do you think ?

### LINKED ISSUE #794: Improve ChronicleReaderMain replacement
A customer has a custom chronicle reader for fix logs which descends from ChronicleReaderMain and overrides configureReader, fixLogRenderer, and options

We should add provisions to obtain equivalent functionality with the new implementations.

minborg:
This PR reverts the deprecation of `ChronicleReaderMain` which is overridable and widely used. Code is deduplicated by letting the `InternalReaderMain` class simply inherit from `ChronicleReaderMain`.

Generally, I think we should break out main classes and similar utility features from the Queue library over time and have a separate repository for this type of functionality. The Queue library should provide as little as possible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -306,6 +306,7 @@
                         </goals>
                         <configuration>
                             <referenceVersion>5.21ea0</referenceVersion>
+                            <binaryCompatibilityPercentageRequired>99.4</binaryCompatibilityPercentageRequired>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
                         </configuration>
                     </execution>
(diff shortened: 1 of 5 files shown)


JerryShea:
Please ignore my commits - I made them to the wrong branch by mistake and have reversed

minborg:
What is your opinion @JerryShea? Can we merge this PR now?

RobAustin:
whats happening with this ? @JerryShea @minborg , can we merge it or close it

minborg:
I will merge this PR on Monday the 18:th at noon (BST) @JerryShea unless you have more comments.

Issue #788, opened 2020-12-30
closed tailer can be re-acquired


JerryShea:
See also #743 and #787

JerryShea:
thanks @nicktindall !

PR #813, opened 2021-03-15
Create a new tailer on acquire if the existing one is closed

nicktindall:
Fixes https://github.com/OpenHFT/Chronicle-Queue/issues/788

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -229,7 +229,14 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
     @Deprecated(/* to be removed in x.23 */)
     @NotNull
     StoreTailer acquireTailer() {
-        return ThreadLocalHelper.getTL(tlTailer, this, q -> new StoreTailer(q, q.pool));
+        StoreTailer tl = ThreadLocalHelper.getTL(tlTailer, this, q -> new StoreTailer(q, q.pool));
+
+        if (tl.isClosing()) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -229,7 +229,14 @@ protected CycleCalculator cycleCalculator(ZoneId zoneId) {
     @Deprecated(/* to be removed in x.23 */)
     @NotNull
     StoreTailer acquireTailer() {
-        return ThreadLocalHelper.getTL(tlTailer, this, q -> new StoreTailer(q, q.pool));
+        StoreTailer tl = ThreadLocalHelper.getTL(tlTailer, this, q -> new StoreTailer(q, q.pool));
```
nicktindall, line 233: I notice a lot of the other methods (including `acquireAppender` and `createTailer`) throw an exception if the queue is closed. Should we do it here for consistency? or is there a use case for getting a tailer on a closed queue?

JerryShea: Yes - good spot - we should check for closed

nicktindall: I added the check.

Previously:
- if `acquireTailer` was called for the first time on a closed Queue it would throw `ClosedIllegalStateException` due to something downstream of the `StoreTailer` constructor.
- if `acquireTailer` was called for a subsequent time on a closed Queue it would not throw

So the change just means you'll now get that error regardless if its the first or subsequent call to `acquireTailer` on a closed queue.

PR #815, opened 2021-03-16
Fix some issues with double buffering, add some specific tests to guard against regression

### LINKED ISSUE #799: net.openhft.chronicle.queue.impl.single.TestMethodWriterWithThreads for doubleBuffer=true
It is working now for doubleBuffer=false

nicktindall:
Fixed #799

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -356,9 +356,11 @@ class StoreAppender extends AbstractCloseable
 
             // sets the writeLimit based on the safeLength
             openContext(metaData, safeLength);
+
+            // there is nothing to read.
+            wire.bytes().readPosition(wire.bytes().writePosition());
         }
-        // there is nothing to read.
-        wire.bytes().readPosition(wire.bytes().writePosition());
+
         return writeContext;
     }
 
@@ -366,7 +368,7 @@ class StoreAppender extends AbstractCloseable
     public DocumentContext acquireWritingDocument(boolean metaData) {
         if (!DISABLE_THREAD_SAFETY)
             this.threadSafetyCheck(true);
-        if (wire != null && writeContext.isOpen() && writeContext.chainedElement())
+        if (writeContext.wire != null && writeContext.isOpen() && writeContext.chainedElement())
             return writeContext;
         return writingDocument(metaData);
     }
@@ -834,6 +836,7 @@ class StoreAppender extends AbstractCloseable
                     if (buffered) {
                         writeBytes(wire.bytes());
                         unlock = false;
+                        wire.clear();
                     } else {
                         writeBytesInternal(wire.bytes(), metaData);
                         wire = StoreAppender.this.wire;
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -357,7 +357,7 @@ class StoreAppender extends AbstractCloseable
             // sets the writeLimit based on the safeLength
             openContext(metaData, safeLength);
 
-            // there is nothing to read.
+            // Move readPosition to the start of the context. i.e. readRemaining() == 0
             wire.bytes().readPosition(wire.bytes().writePosition());
         }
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -356,17 +356,19 @@ public DocumentContext writingDocument(final boolean metaData) {
 
             // sets the writeLimit based on the safeLength
             openContext(metaData, safeLength);
+
+            // there is nothing to read.
```
nicktindall: TODO: Still need to improve this comment

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -356,17 +356,19 @@ public DocumentContext writingDocument(final boolean metaData) {
...
+
         return writeContext;
     }
 
     @Override
     public DocumentContext acquireWritingDocument(boolean metaData) {
         if (!DISABLE_THREAD_SAFETY)
             this.threadSafetyCheck(true);
-        if (wire != null && writeContext.isOpen() && writeContext.chainedElement())
+        if (writeContext.wire != null && writeContext.isOpen() && writeContext.chainedElement())
```
nicktindall, line 371: This broke the chained `methodWriter` writes when double buffering. It mean `count` never got to zero so the buffer was never written back.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -834,6 +836,7 @@ public void close(boolean unlock) {
                     if (buffered) {
                         writeBytes(wire.bytes());
                         unlock = false;
+                        wire.clear();
```
nicktindall, line 839: When the buffer was used the second time, the previous calls were still in it.

src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleMultiThreadStressDoubleBufferTest.java
```diff
@@ -1,19 +1,93 @@
...
+                    if (unexpectedValues.contains(expected)) {
+                        unexpectedValues.remove(expected);
+                    } else {
+                        skippedValue.add(expected);
+                    }
+                    if (skippedValue.contains(v)) {
+                        skippedValue.remove(v);
+                    } else {
+                        unexpectedValues.add(v);
+                    }
```
nicktindall, line 72: I'm sure there's a cleverer/more efficient way to track this... I'll come back to it.

src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleMultiThreadStressTest.java
```diff
@@ -477,6 +520,10 @@ private int write(ExcerptAppender appender) {
             try (DocumentContext writingDocument = appender.writingDocument()) {
                 final long documentAcquireTimestamp = System.nanoTime();
                 value = wrote.getAndIncrement();
+                if (value >= expectedNumberOfMessages) {
+                    writingDocument.rollbackOnClose();
+                    return value;
+                }
```
nicktindall, line 533: I needed to do this because threads can open `DocumentContext`s concurrently when double buffering is enabled, so in those circumstances they'd sometimes write more than `expectedNumberOfMessages`

nicktindall: Combine that with being potentially out of order and it becomes very hard to check on the reader side.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -856,12 +859,17 @@ public void close(boolean unlock) {
...
+                assert wire != StoreAppender.this.wire;
+                wire.clear();
+            } else {
+                // zero out all contents...
+                for (long i = positionOfHeader; i <= wire.bytes().writePosition(); i++)
+                    wire.bytes().writeByte(i, (byte) 0);
+                long lastPosition = StoreAppender.this.lastPosition;
+                position0(lastPosition, lastPosition);
+                ((AbstractWire) wire).forceNotInsideHeader();
+            }
```
nicktindall, line 872: This isn't pretty, but it gets the job done. Previously `doRollback` assumed `StoreAppenderContext.wire == StoreAppender.wire`

src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleMultiThreadStressTest.java
JerryShea: maybe add a comment to that code explaining why?

JerryShea (approved):

PR #819, opened 2021-04-07
make cycle handling consistent

JerryShea:
Making handle of cycles consistent. I don't necessarily think this will change behaviour but keen to get all eyes on this


JerryShea:
See also https://github.com/OpenHFT/Chronicle-Queue/issues/820

minborg:
Looks consistent to me.

Issue #822, opened 2021-04-08
Cannot read from queue if first message is metadata


JerryShea:
If the 1st message written to a queue is metadata then that message can not be read from the queue. If a subsequent non-metadata message is written then that can be read. See net.openhft.chronicle.queue.ChronicleQueueIndexTest#writeReadMetadata

hft-team-city:
Released in [Chronicle-Queue-5.21ea43](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.21ea43), [BOM-2.21ea110](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.21ea110)

PR #824, opened 2021-04-08
https://github.com/OpenHFT/Chronicle-Queue/issues/822

JerryShea:
Index entry required if first item in a queue is metadata

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -851,7 +851,7 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
                     // file not found.
                     if (store == null)
                         break;
-                    if (store.writePosition() == 0 && !store.file().delete() && store.file().exists()) {
+                    if (store.writePosition() == 0 && store.file().exists()) {
                         // couldn't delete? Let's try writing EOF
                         // if this blows up we should blow up too so don't catch anything
                         MappedBytes bytes = store.bytes();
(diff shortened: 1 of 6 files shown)


src/test/java/net/openhft/chronicle/queue/ChronicleQueueIndexTest.java
```diff
@@ -292,4 +293,157 @@ public void writeReadMetadata() {
             }
         }
     }
+
+    // https://github.com/OpenHFT/Chronicle-Queue/issues/822
+    private void driver0(String[] strings, boolean[] meta, long millis) {
+
+        assert (strings.length == meta.length);
+
+        File tmp = new File("/dev/shm/Issue882");
```
JerryShea: not cross platform - I would use getTmpDir

JerryShea: other than this minor change this all LGTM although I would not rely on just me ;)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -851,7 +851,7 @@ void cleanupStoreFilesWithNoData() {
                     // file not found.
                     if (store == null)
                         break;
-                    if (store.writePosition() == 0 && !store.file().delete() && store.file().exists()) {
+                    if (store.writePosition() == 0 && store.file().exists()) {
                         // couldn't delete? Let's try writing EOF
```
JerryShea: we should change the comment too ;)

src/test/java/net/openhft/chronicle/queue/ChronicleQueueIndexTest.java
rogersimmons: Thanks @JerryShea  - yes, only intended while testing locally. Swapped to getTmpDir()

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
rogersimmons: updated

chronicle-dev (approved):

PR #828, opened 2021-04-12
Review appender write bytes

JerryShea:
@nicktindall

I have added some additional tests and sprinkled some TODOs around for discussion

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
@@ -189,10 +189,9 @@ public class InternalAppenderWriteBytesTest extends ChronicleQueueTestBase {
             // here we try and write to previous cycle file. We will overwrite the EOF in doing so
             expectException("Incomplete header found at pos: 33048: c0000000, overwriting");
             ((InternalAppender) appender).writeBytes(nextIndexInFirstCycle, test1);
-            // TODO: if we are going to support this use case we need to fix EOF
             Assert.assertFalse(hasEOF(q, firstCycle));
 
-            // TODO: as EOF has not been written, the queue is now sick. For now we have to manually fix
+            // we have to manually fix. This is done by CQE at the end of backfilling
             appender.normaliseEOFs();
 
             ExcerptTailer tailer = q.createTailer();
--- a/src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
@@ -12,6 +12,7 @@ import net.openhft.chronicle.wire.DocumentContext;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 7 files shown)


src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
```diff
@@ -1,80 +1,129 @@
...
+        Bytes result = Bytes.elasticHeapByteBuffer();
+        try (SingleChronicleQueue q = SingleChronicleQueueBuilder.binary(getTmpDir()).timeProvider(() -> 0).build()) {
+            ExcerptAppender appender = q.acquireAppender();
+            appender.writeBytes(test);
+        }
+
+        try (SingleChronicleQueue q = SingleChronicleQueueBuilder.binary(getTmpDir()).timeProvider(() -> 0).build()) {
+            ExcerptAppender appender = q.acquireAppender();
+            // TODO: this overwrites because the appender's wire's headerNumber is positioned at the start
+            ((InternalAppender) appender).writeBytes(0, Bytes.from("HELLO WORLD"));
```
JerryShea, line 96: this is a defect that we need to fix

src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
```diff
@@ -113,17 +153,64 @@ public void testJumpingAMessageThrowsAIllegalStateException() {
...
+            long nextIndexInFirstCycle = appender.lastIndexAppended() + 1;
+            int firstCycle = q.rollCycle().toCycle(nextIndexInFirstCycle);
+
+            timeProvider.advanceMillis(TimeUnit.SECONDS.toMillis(65));
+            appender.writeBytes(test2);
+
+            Assert.assertTrue(hasEOF(q, firstCycle));
+            // here we try and write to previous cycle file. We will overwrite the EOF in doing so
+            expectException("Incomplete header found at pos: 33048: c0000000, overwriting");
+            ((InternalAppender) appender).writeBytes(nextIndexInFirstCycle, test1);
```
JerryShea, line 191: this does work currently i.e. we can add to the previous cycle. If we want to support this as a use case we will need to re-write the EOF at the end, if we overwrote it

JerryShea: this feature could be useful to keep in case e.g. the pretoucher has rolled the queue already. Although if the queue has rolled and there is data in the next cycle then we probably don't want to allow this

src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
JerryShea: @peter-lawrey just reminding you that you put your hand up for this ;)

src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
JerryShea: We are happy with current behaviour. javadoc now reflects this test

PR #830, opened 2021-04-13
Implement lastIndex on ChronicleQueue

nicktindall:
As discussed last night, I've added an implementation of `lastIndex` for use when we are handshaking replication.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
@@ -212,6 +212,19 @@ public interface ChronicleQueue extends Closeable {
     long firstIndex();
 
     /**
+     * Returns the index of the last non-metadata excerpt written to this ChronicleQueue,
+     * or -1 if the queue is empty.
+     * <p>
+     * The value returned by this method will not be reliable in the event the queue
+     * is being written to concurrently, as subsequent excerpts may have been written
+     * by the time it is returned.
+     *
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
```diff
@@ -211,6 +211,19 @@ default LongValue indexForId(String id) {
      */
     long firstIndex();
 
+    /**
+     * Returns the highest valid index available for this ChronicleQueue, or -1 if
```
JerryShea: valid == not metadata?

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
```diff
@@ -3611,4 +3611,49 @@ public void testReadUsingReadOnly() {
...
+                try (DocumentContext documentContext = appender.writingDocument(true)) {
+                    documentContext.wire().write().text("Hello!");
+                }
+            }
+            assertEquals(-1, queue.lastIndex());
+        }
+    }
+
+    @Test
+    public void lastIndexShouldReturnNegativeOneForEmptyPretouchedQueue() {
```
JerryShea, line 3650: nice

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
nicktindall: Fixed

PR #841, opened 2021-04-26
Stop locking whole file for "exclusive" lock in single table store

nicktindall:
In order that we can hold long-running fine-grained locks on values (e.g the lock that indicates a replicator is running) we need to move away from attempting to lock the whole file to mutex structural changes to the table store (i.e. acqure value, writing first header).
This change locks `Long.MAX_VALUE-1 -> Long.MAX_VALUE` to implement the "exclusive" lock. This shouldn't overlap with any other locks in the table store until we have a `SingleTableStore` so large we'd need something other than FileChannel to deal with it.

**NOTE** Any other implementations of SingleTableStore need to be updated to match this (i.e. the C++ version)

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/table/SingleTableStore.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/table/SingleTableStore.java
@@ -50,6 +50,13 @@ import static net.openhft.chronicle.core.util.Time.sleep;
 
 public class SingleTableStore<T extends Metadata> extends AbstractCloseable implements TableStore<T> {
     public static final String SUFFIX = ".cq4t";
+    private static final int EXCLUSIVE_LOCK_SIZE = 1;
+    /**
+     * We need to be able to acquire an "exclusive" lock while fine-grained long-running locks are being held.
+     * For this reason the "exclusive" lock doesn't lock the whole file, but as long as everyone agrees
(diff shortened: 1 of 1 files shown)


JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/table/SingleTableStore.java
```diff
@@ -50,6 +50,13 @@
 
 public class SingleTableStore<T extends Metadata> extends AbstractCloseable implements TableStore<T> {
     public static final String SUFFIX = ".cq4t";
+    private static final int EXCLUSIVE_LOCK_SIZE = 1;
+    /**
+     * We need to be able to acquire an "exclusive" lock while fine-grained long-running locks are being held.
+     * For this reason the "exclusive" lock doesn't lock the whole file, but as long as everyone agrees
+     * on what section constitutes an "exclusive" lock this should be fine.
+     */
+    private static final long EXCLUSIVE_LOCK_START = Long.MAX_VALUE - EXCLUSIVE_LOCK_SIZE;
```
JerryShea, line 59: I like this - all the other locks around long values will all have low offsets, and it is clear from looking at the offsets that the exclusive lock is a different beast. Nice intentional code :)

PR #843, opened 2021-04-27
Make createAppender blocking pluggable

nicktindall:
This PR just moves to using a `Condition` to block the creation of a new appender.

Currently we await on the QueueLock to be unlocked before a new appender is returned.

This change allows the Condition on which we wait to be configured. So the existing behaviour (waiting on the QueueLock) can be preserved until .22, but we can optionally provide a more sophisticated condition for those use cases where that's required.

The Condition interface is a bit of overkill, but its nice to use a standard interface here.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
@@ -30,7 +30,10 @@ import static net.openhft.chronicle.core.Jvm.warn;
 
 /**
  * Implements queue lock via TableStore mechanism.
+ *
+ * @deprecated To be removed in .22
  */
+@Deprecated
 public class TSQueueLock extends AbstractTSQueueLock implements QueueLock {
 
     private static final String LOCK_KEY = "chronicle.queue.lock";
@@ -152,6 +155,11 @@ public class TSQueueLock extends AbstractTSQueueLock implements QueueLock {
                 warn().on(getClass(), "Queue lock was locked by another thread, current-thread-tid=" + tid + ", lock value=" + value+", this lock was not removed.");
             }
         }
+    }
+
+    @Override
+    public boolean isLocked() {
+        return lockedBy() != UNLOCKED;
     }
 
     private boolean isLockHeldByCurrentThread(long tid) {
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
@@ -31,6 +31,7 @@ import net.openhft.chronicle.core.util.StringUtils;
 import net.openhft.chronicle.queue.*;
 import net.openhft.chronicle.queue.impl.RollingChronicleQueue;
 import net.openhft.chronicle.threads.NamedThreadFactory;
+import net.openhft.chronicle.threads.TimeoutPauser;
 import net.openhft.chronicle.wire.*;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
@@ -51,6 +52,8 @@ import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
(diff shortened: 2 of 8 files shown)


src/main/java/net/openhft/chronicle/queue/impl/ThreadGate.java
```diff
@@ -0,0 +1,9 @@
+package net.openhft.chronicle.queue.impl;
+
+/**
+ * A "gate" at which threads wait for some condition
+ */
+public interface ThreadGate {
+
+    void waitFor();
+}
```
nicktindall: I looked but couldn't see a standard interface from `java.util.concurrent` to use here. If there was I'd be very open to using it. Also happy to bikeshed about the naming here, maybe `...Barrier` or something else is more appropriate?

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+            return fileLock == null;
+        } catch (OverlappingFileLockException e) {
+            return true;
+        } catch (IOException e) {
+            throw new IllegalStateException("Error checking isLocked " + lockName, e);
+        }
+    }
+}
+
```
nicktindall, line 71: I'm not sure if I should instead give this a similar API to `WriteLock`. Its a little bit different to that, in that only the holder of the lock can unlock it (you need the FileLock object to do that).

nicktindall: It would be a bit weird because only the INSTANCE that locked() the lock would be able to unlock, even if the process was the one holding the lock. Unless I store the lock in a static field on the class? Then you'd probably have to synchronize the lock and unlock methods.

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
```diff
@@ -30,7 +30,10 @@
 
 /**
  * Implements queue lock via TableStore mechanism.
+ *
+ * @deprecated To be removed in .22
  */
+@Deprecated
 public class TSQueueLock extends AbstractTSQueueLock implements QueueLock {
```
JerryShea, line 37: ah so queueLock is the only thing that uses this?

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
```diff
@@ -0,0 +1,71 @@
...
+    public FileLock lock() {
+        return lock(lockLongReference, lockName);
+    }
+
+    public static FileLock lock(BinaryLongReference binaryLongReference, String lockName) {
+        FileLock lock;
+        try {
+            lock = binaryLongReference.lock(false);
+            if (lock != null) {
+                // This process acquired the replicator lock
```
JerryShea, line 35: so this means the other process is dead?

JerryShea: or no process has ever locked

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
```diff
@@ -0,0 +1,71 @@
...
+                if (binaryLongReference.getVolatileValue() != PID) {
```
JerryShea, line 36: so we allow re-entrancy

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static boolean isLocked(BinaryLongReference lockReference, String lockName) {
+        try {
+            // When testing isLocked try and acquire a shared lock so we don't mistake
+            // others testing isLocked for the actual holder
+            final FileLock fileLock = lockReference.tryLock(true);
+            if (fileLock != null) {
+                fileLock.release();
+            }
+            return fileLock == null;
```
JerryShea, line 63: so this means that we have the file lock so we don't need to test the value in the LongValue? Maybe we should assert it is == PID?

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
nicktindall: Either of those, you can't acquire the lock if its being held by another live process so a lock being returned means it was acquired.

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
nicktindall: No, if the current (or any) process holds the lock, `binaryLongReference.lock` will throw an `OverlappingFileLockException`

src/main/java/net/openhft/chronicle/queue/util/NamedTableStoreLock.java
nicktindall: No, see above, the lock isn't re-entrant at the OS level so we won't get in here if the lock was being held by anyone

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
nicktindall: Yup

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
```diff
@@ -3657,4 +3659,77 @@ public void lastIndexShouldReturnNegativeOneForEmptyPretouchedQueue() {
...
     }
+
+    @Test
+    public void shouldWaitForConditionWhenAcquiringAppender() throws TimeoutException {
+        File tmpDir = getTmpDir();
+        AtomicBoolean gotAppender = new AtomicBoolean(false);
+        Semaphore acquireAppenderLock = new Semaphore(0);
+        try (final SingleChronicleQueue queue = SingleChronicleQueueBuilder.single(tmpDir)
+                .wireType(wireType)
+                .acquireAppenderConditionCreator((ts, ql) -> new SemaphoreCondition(acquireAppenderLock))
```
JerryShea: nice

JerryShea (approved):

rogersimmons (approved):

PR #854, opened 2021-05-20
Add -z option to print micro timestamps in local timezone, fixes #832

### LINKED ISSUE #832: ChronicleReader to optionally convert date/times in output into local TZ
When reading date times encoded using e.g. `MicroTimestampLongConverter` using `-r` they are always displayed in UTC. 

It may be challenging to satisfy this requirement so need to discuss the cost/benefit

nicktindall:


src/main/java/net/openhft/chronicle/queue/ChronicleReaderMain.java
```diff
@@ -136,6 +137,9 @@ protected void configureReader(final ChronicleReader chronicleReader, final Comm
         if (commandLine.hasOption('s')) {
             chronicleReader.suppressDisplayIndex();
         }
+        if (commandLine.hasOption('z')) {
+            System.setProperty("mtlc.zoneId", ZoneId.systemDefault().toString());
+        }
```
nicktindall, line 142: I wanted to leave this in ChronicleReaderMain not in ChroniclReader because its a global change

JerryShea (approved):

PR #856, opened 2021-05-21
Make write lock pauser thread-safe, fixes #844

### LINKED ISSUE #844: Default TimedPauser in WriteLock is not thread safe
If `pauserSupplier` isn't specified in `SingleChronicleQueueBuilder` you get `SingleChronicleQueueBuilder.DefaultPauserSupplier`

https://github.com/OpenHFT/Chronicle-Queue/blob/482cfe3c15fbb71a42ad3dd83c60ebe74654e9c2/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java#L1121

Which returns a new `TimoutPauser` each time its called. This is used for the pauser for `appendLock`, `writeLock` and `queueLock`, all of which are hittably by multiple threads, but TimeoutPauser isn't thread-safe.

nicktindall:


src/main/java/net/openhft/chronicle/queue/impl/table/AbstractTSQueueLock.java
```diff
@@ -38,15 +39,15 @@
     protected final boolean dontRecoverLockTimeout = Jvm.getBoolean("queue.dont.recover.lock.timeout");
 
     protected final LongValue lock;
-    protected final TimingPauser pauser;
+    protected final CleaningThreadLocal<TimingPauser> pauser;
```
nicktindall: Not sure if I should just use normal ThreadLocal here considering none of the TimingPausers have any requirement for cleanup.

JerryShea: Yep good question. Might be more intentional to use a normal TL?

src/main/java/net/openhft/chronicle/queue/impl/table/AbstractTSQueueLock.java
```diff
@@ -38,15 +39,15 @@
...
     protected final File path;
     protected final TableStore tableStore;
     private final String lockKey;
 
-    public AbstractTSQueueLock(final String lockKey, final TableStore<?> tableStore, final Supplier<TimingPauser> pauser) {
+    public AbstractTSQueueLock(final String lockKey, final TableStore<?> tableStore, final Supplier<TimingPauser> pauserSupplier) {
```
JerryShea, line 46: And we don't even need to change the API! cool

JerryShea (approved):

PR #898, opened 2021-08-06
Manage nested writing document calls counter

### LINKED ISSUE #896: StoreAppender can bypass write lock after an exception is thrown in writingDocument()
When a RuntimeException is thrown, the counter we use to keep track of nested calls to `StoreAppender.writingDocument()` can end up being out of sync leaving callers able to get a DocumentContext configured from a previous write. This can lead to queue corruption due to over-writing of already-written queue entries and concurrent writes occurring due to the write lock acquisition being bypassed.

nicktindall:
Fixes #896

Also use InterruptedRuntimeException instead of IllegalStateException and stop clearing interrupted flag in interrupted locks

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
@@ -18,6 +18,7 @@
 package net.openhft.chronicle.queue.impl.single;
 
 import net.openhft.chronicle.core.StackTrace;
+import net.openhft.chronicle.core.threads.InterruptedRuntimeException;
 import net.openhft.chronicle.queue.impl.TableStore;
 import net.openhft.chronicle.queue.impl.table.AbstractTSQueueLock;
 import net.openhft.chronicle.threads.TimingPauser;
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -64,8 +65,8 @@ public void lock() {
             long start = System.currentTimeMillis();
             while (!lock.compareAndSwapValue(UNLOCKED, PID)) {
                 // add a tiny delay
-                if (Thread.interrupted())
-                    throw new IllegalStateException("Interrupted for the lock file:" + path);
+                if (Thread.currentThread().isInterrupted())
```
JerryShea, line 68: should we `Thread.currentThread().interrupt(); ?`

JerryShea (approved):
fantastic

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
nicktindall: No
> The interrupted status of the thread is unaffected by this method.

From https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#isInterrupted--

JerryShea: Oh yes of course. It is friday afternoon after all ;)

PR #900, opened 2021-08-09
Support binary search in the presence of excerpts that can't be compared

nicktindall:
In implementing the start-from-timestamp functionality in chronicle reader, we ran into the issue in services that there are queue excerpts that contain only history entries and no "payload". This meant that binary search was unable to continue. This PR changes the binary search to be tolerant of "sparse" data. Comparators can optionally throw `NotComparableException` when given an excerpt that can't be compared, and the binary search will walk forward until it finds an entry that can be compared.

The assumption here is that non-comparable excerpts are rare and don't appear in long sequences. The less true that assumption is, the worse this search will perform and may perform worse than a linear search in extreme cases.

An optimisation would be to make more effort to find the "middle" comparable element, so trying elements in distance-from-current-middle order (i.e. in both directions) rather than just moving the "middle" to the right like we do. That would be an optimisation we could add easily if it was deemed worthwhile.

With a comparator that never throws `NotComparableException`, the behaviour should be identical to the previous implementation.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/BinarySearch.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/BinarySearch.java
@@ -55,6 +55,7 @@ public enum BinarySearch {
                                               @NotNull Comparator<Wire> c,
                                               @NotNull ExcerptTailer tailer,
                                               @NotNull final ChronicleQueue queue) {
+        long readPosition = key.bytes().readPosition();
 
         final Iterator<Long> iterator = cycles.iterator();
         if (!iterator.hasNext())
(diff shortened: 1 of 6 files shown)


JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/TestBinarySearch.java
```diff
@@ -14,13 +14,16 @@
...
 import java.text.ParseException;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Comparator;
+import java.util.*;
+
+import static org.junit.Assert.assertTrue;
 
 @RunWith(Parameterized.class)
 public class TestBinarySearch extends ChronicleQueueTestBase {
```
JerryShea: please add @RequiredForClient annotation

JerryShea (approved):

PR #914, opened 2021-09-29
Improve exception message when store not found for cycle, assert TableDirectoryListing is not used in readOnly queues

nicktindall:


JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -749,7 +749,7 @@ private long approximateLastCycle2(int lastCycle) throws StreamCorruptedExceptio
                 lastCycle, queue.epoch(), false, this.store);
         this.setCycle(lastCycle);
         if (wireStore == null)
-            throw new IllegalStateException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed?");
+            throw new IllegalStateException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? lastCycle=" + lastCycle + ", queue=" + queue.fileAbsolutePath());
```
JerryShea: we already output lastCycle

nicktindall: hehe good point...

nicktindall: I copied that from below

src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
```diff
@@ -31,11 +33,13 @@
             final @NotNull TableStore<?> tableStore,
             final Path queuePath,
             final ToIntFunction<File> fileToCycleFunction,
-            final boolean readOnly) {
+            final boolean readOnly,
```
JerryShea: you could drop readOnly as an argument and assert !tableStore.readOnly()

nicktindall: Yep, love it

PR #921, opened 2021-10-07
Expose forceUnlockIfProcessIsDead and isLockedByCurrentProcess via the WriteLock interface

nicktindall:
Just so we don't need to downcast to do this that whole lock hierarchy could probably do with a review, I wonder how much if it is still necessary

Also changed so that ChronicleQueue subclasses don't have to re-implement the createAppenderCondition blocking when overriding appender creation

And publish test sources, to go with the test jar we already publish

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,6 +229,22 @@
                     <scmBranch>gh-pages</scmBranch>
                 </configuration>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>3.2.1</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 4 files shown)


JerryShea (approved):

pom.xml
```diff
@@ -229,6 +229,22 @@
...
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
+                        <goals>
+                            <goal>jar</goal>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
nicktindall, line 246: We publish a test-jar from CQ and we reuse some of the tests in CQE, so it's nice to have the source available when you're debugging

minborg (approved):

PR #925, opened 2021-10-12
Optimize table directory refresh

### LINKED ISSUE #922: Reduced unwanted Object creation
This relates to the following support issue -- https://support.chronicle.software/staff/ticket/422

The following is from a YourKit Dump

image

glukos:
The fix is based on the fact that lexicographical order of queue files matches the chronological order. It allows to avoid parsing every file name, which is quite expensive. We still generate an array of strings for every refresh, but it demonstrates surprisingly low pressure, probably due to built-in string deduplication (since Java 8).
I have also tried approaches with `Files#walkFileTree` or directory streams, they appeared to be very garbag-y as well.
Allocation stats for 1 minute run are below, the fix decreases total amount of allocated space from 31M to 56K.

Before the fix:
before_fix_alloc
before_fix_gc
After the fix:
after_fix_alloc
after_fix_gc

The difference was measured in synthetic scenario:
- RollCycles.MINUTELY
- ~5k queue files
- forceDirectoryListingRefreshIntervalMs is forced to 1s
- jvm.resource.tracing=false
- assertions disabled
- one record is appended to the queue every 1s

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/FileSystemDirectoryListing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/FileSystemDirectoryListing.java
@@ -5,17 +5,19 @@ import net.openhft.chronicle.core.io.SimpleCloseable;
 import java.io.File;
 import java.util.function.ToIntFunction;
 
+import static net.openhft.chronicle.queue.impl.single.TableDirectoryListing.*;
+
 final class FileSystemDirectoryListing extends SimpleCloseable implements DirectoryListing {
     private final File queueDir;
-    private final ToIntFunction<File> fileToCycleFunction;
+    private final ToIntFunction<String> fileNameToCycleFunction;
     private int minCreatedCycle = Integer.MAX_VALUE;
     private int maxCreatedCycle = Integer.MIN_VALUE;
     private long lastRefreshTimeMS;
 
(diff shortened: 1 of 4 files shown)


JerryShea (approved):

nicktindall (approved):
This is a nice fix, are we sure it's OK to assume the correlation between lexicographical ordering and chronological ordering? is that part of the roll cycle specification, are there tests to guard against that assumption being broken (e.g. by a client implementing their own roll cycle, if that's a thing that happens)?

JerryShea:
> This is a nice fix, are we sure it's OK to assume the correlation between lexicographical ordering and chronological ordering? is that part of the roll cycle specification, are there tests to guard against that assumption being broken (e.g. by a client implementing their own roll cycle, if that's a thing that happens)?

that's a good point. It is certainly true for `net.openhft.chronicle.queue.RollCycles` but I guess may not be for a custom RollCycle

minborg:
Maybe we should require correlation between lexicographical ordering and chronological ordering as a requirement for any RollCycle implementation in the JavaDoc?

JerryShea:
+1 Yes - we should at least document this assumption in RollCycle javadoc

glukos:
Merged under https://github.com/OpenHFT/Chronicle-Queue/issues/922, I've added a note in JavaDoc

PR #926, opened 2021-10-13
Recover from _some_ missing files scenarios more gracefully, improve testing

### LINKED ISSUE #913: When first and last cycle are out of date, should we try refreshing them and retrying before we throw?
In `StoreTailer.toEnd`, we get the `lastCycle` which is only checked against the file listing every 60s. In the event we've got `lastCycle` then failed to retrieve it from the `SingleChronicleQueueStore` is it worth forcing a `directoryListing.refresh(true)` and retrying before throwing an exception?

nicktindall:
Fixes #913 (somewhat)

This PR fixes some of the low-hanging-fruit places where we used to choke on a missing file. Basically just repeating the pattern of "refresh directory listing and retry once when a file that we expect to be there is not".

Still lots of things will be broken under this scenario but this PR makes it a tiny bit more resilient, while being, I think, a low-risk change.

I won't be offended if this is rejected, I was ambivalent about the change, because it's not a proper fix. But perfect is the enemy of good.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/MissingStoreFileException.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/MissingStoreFileException.java
@@ -1,0 +1,10 @@
+package net.openhft.chronicle.queue.impl.single;
+
+/**
+ * Thrown when a store file we expect to be present is missing (probably because it was deleted)
+ */
+public class MissingStoreFileException extends IllegalStateException {
+    public MissingStoreFileException(String s) {
+        super(s);
+    }
+}
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -687,9 +687,7 @@ class StoreTailer extends AbstractCloseable
         return scanResult;
     }
 
-    @NotNull
-    @Override
-    public final ExcerptTailer toStart() {
+    private ExcerptTailer doToStart() {
         assert direction != BACKWARD;
         final int firstCycle = queue.firstCycle();
         if (firstCycle == Integer.MAX_VALUE) {
@@ -700,9 +698,10 @@ class StoreTailer extends AbstractCloseable
         if (firstCycle != this.cycle) {
             // moves to the expected cycle
             final boolean found = cycle(firstCycle);
-            assert found || store == null;
             if (found)
                 state = FOUND_IN_CYCLE;
+            else if (store != null)
+                throw new MissingStoreFileException("Missing first store file cycle=" + firstCycle);
         }
         index(queue.rollCycle().toIndex(cycle, 0));
 
@@ -713,6 +712,17 @@ class StoreTailer extends AbstractCloseable
             address = wire.bytes().addressForRead(0);
         }
         return this;
+    }
+
+    @NotNull
+    @Override
+    public final ExcerptTailer toStart() {
+        try {
+            return doToStart();
+        } catch (MissingStoreFileException e) {
+            queue.refreshDirectoryListing();
+            return doToStart();
+        }
     }
 
     private boolean moveToIndexInternal(final long index) {
@@ -749,7 +759,7 @@ class StoreTailer extends AbstractCloseable
                 lastCycle, queue.epoch(), false, this.store);
         this.setCycle(lastCycle);
         if (wireStore == null)
-            throw new IllegalStateException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
+            throw new MissingStoreFileException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
 
         if (this.store != wireStore) {
             releaseStore();
@@ -769,7 +779,7 @@ class StoreTailer extends AbstractCloseable
                 lastCycle--;
                 try {
                     return approximateLastCycle2(lastCycle);
-                } catch (IllegalStateException e) {
+                } catch (MissingStoreFileException e) {
                     // try again.
                 }
             }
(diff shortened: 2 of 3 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -700,9 +698,10 @@ public final ExcerptTailer toStart() {
         if (firstCycle != this.cycle) {
             // moves to the expected cycle
             final boolean found = cycle(firstCycle);
-            assert found || store == null;
             if (found)
                 state = FOUND_IN_CYCLE;
+            else if (store != null)
+                throw new MissingStoreFileException("Missing first store file cycle=" + firstCycle);
```
nicktindall, line 704: Converted the above assertion (which fails when the file is deleted) into a retry then throw if not resolved

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -715,6 +714,17 @@ public final ExcerptTailer toStart() {
...
+    @NotNull
+    @Override
+    public final ExcerptTailer toStart() {
+        try {
+            return doToStart();
+        } catch (MissingStoreFileException e) {
+            queue.refreshDirectoryListing();
+            return doToStart();
+        }
+    }
```
nicktindall, line 726: Wrap `toStart` to retry once in the specific case that the first cycle is not there and we think it should be

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -749,7 +759,7 @@ private long approximateLastCycle2(int lastCycle) throws StreamCorruptedExceptio
                 lastCycle, queue.epoch(), false, this.store);
         this.setCycle(lastCycle);
         if (wireStore == null)
-            throw new IllegalStateException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
+            throw new MissingStoreFileException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
```
nicktindall, line 762: `MissingStoreFileException` is more specific, and extends `IllegalStateException` so shouldn't change the contract with the outside world

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -769,7 +779,7 @@ private long approximateLastCycle2(int lastCycle) throws StreamCorruptedExceptio
                 lastCycle--;
                 try {
                     return approximateLastCycle2(lastCycle);
-                } catch (IllegalStateException e) {
+                } catch (MissingStoreFileException e) {
```
nicktindall, line 782: I _think_ it's better to only keep looking on a missing file exception, otherwise we'd do it for, e.g. a `ClosedIllegalStateException` which absolutely can be thrown by that method

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -837,7 +847,17 @@ public ExcerptTailer toEnd() {
...
+    }
+
+    private ExcerptTailer callOptimizedToEnd() {
+        try {
+            return optimizedToEnd();
+        } catch (MissingStoreFileException e) {
+            Jvm.warn().on(StoreTailer.class, "Missing store file detected, refreshing directory listing and retrying");
+            queue.refreshDirectoryListing();
+            return optimizedToEnd();
+        }
```
nicktindall, line 859: Retry `optimizedToEnd` once in the specific case we encounter a missing store file

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -888,7 +908,7 @@ private ExcerptTailer optimizedToEnd() {
                     lastCycle, queue.epoch(), false, this.store);
             this.setCycle(lastCycle);
             if (wireStore == null)
-                throw new IllegalStateException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
+                throw new MissingStoreFileException("Store not found for cycle " + Long.toHexString(lastCycle) + ". Probably the files were removed? queue=" + queue.fileAbsolutePath());
```
nicktindall, line 910: Again, extends `IllegalStateException` so no change to the contract here

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -2,146 +2,272 @@
...
-            readText(excerptTailer, 5, "test2");
+            // should be at correct index
+            assertEquals(Long.toHexString(secondCycle.firstIndex), Long.toHexString(tailer.toStart().index()));
+        }
+    }
 
-            assertEquals(indexAfter15Records, excerptTailer.index() - 1);
+    @Ignore("still doesn't work")
+    @Test
+    public void tailerToStartFromStartWorksInFaceOfDeletedStoreFile() throws IOException {
```
nicktindall, line 85: `toStart` is only resilient when it results in a load of a new store for some reason, so if we're already at the start it's still broken. I'm hesitant to fix that because it's potentially a bigger change.

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/MissingStoreFileException.java
```diff
@@ -0,0 +1,10 @@
+package net.openhft.chronicle.queue.impl.single;
+
+/**
+ * Thrown when a store file we expect to be present is missing (probably because it was deleted)
+ */
+public class MissingStoreFileException extends IllegalStateException {
```
JerryShea, line 6: like

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
JerryShea: @peter-lawrey would be good to get your input

glukos (approved):
Looks like a good temporary solution until we come closer to the proper fix of #913.

PR #931, opened 2021-10-18
SCQIndexing optimize allocations

### LINKED ISSUE #927: SCQIndexing: passing newLogArrayValuesHolder as method reference triggers allocation every time the entry is indexed
Proof:
image

glukos:


RobAustin (approved):

RobAustin (commented):
I'd be very surprised if this fixes the garbage problem, using a method reference should not create any garbage

glukos:
@RobAustin This is suprising to me as well.
Debugger clearly shows that a new Function instance is passed every time we get into ThreadLocalHelper.getTL.

JProfiled writing to a queue in a loop for 1 minute, before the fix:
lamnda_before_fix_alloc

JProfiled writing to a queue in a loop for 1 minute, after the fix:
lamnda_after_fix_alloc

glukos:
Reproduces on both Java 8 and Java 11.

PR #950, opened 2021-11-16
Refresh directory listing and retry before failing nextCycle due to a missing store file

### LINKED ISSUE #948: Refresh directory listing and retry when a missing file is encountered during nextCycle
When a deleted file is encountered during a call to nextCycle, we throw an IllegalStateException. We could probably handle this more gracefully if we attempted to forcefully refresh the directory listing when this happens and retried once.

The exception is as follows, it could probably be fixed in `net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.nextCycle`
```
java.lang.IllegalStateException: Expected file to exist for cycle: 454664, file: queuefile.cq4. minCycle: 454705, maxCycle: 454705 Available files: [20211115-01LS.cq4]
        at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.nextCycle(SingleChronicleQueue.java:1226) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.WireStorePool.nextCycle(WireStorePool.java:66) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue.nextCycle(SingleChronicleQueue.java:568) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.nextIndexWithNextAvailableCycle(StoreTailer.java:516) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.endOfCycle(StoreTailer.java:341) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.next0(StoreTailer.java:314) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.readingDocument0(StoreTailer.java:215) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.readingDocument(StoreTailer.java:190) ~[chronicle-queue-5.22ea10.jar:?]
        at net.openhft.chronicle.queue.impl.single.StoreTailer.readingDocument(StoreTailer.java:134) ~[chronicle-queue-5.22ea10.jar:?]
```

nicktindall:
Fixes #948

I don't like this fix, and the fact we've implemented similar fixes in a few places makes me think there's a better place to catch this stuff, but the tests indicate it fixes the reported problem.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -1223,7 +1223,7 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
 
                 if (!fileFound) {
                     directoryListing.refresh(true);
-                    throw new IllegalStateException(
+                    throw new MissingStoreFileException(
                             String.format("Expected file to exist for cycle: %d, file: %s.%nminCycle: %d, maxCycle: %d%n" +
                                             "Available files: %s",
                                     currentCycle, currentCycleFile,
@@ -1240,7 +1240,7 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
                 file = tree.get(key);
             }
             if (file == null) {
-                throw new AssertionError("missing currentCycle, file=" + currentCycleFile);
+                throw new MissingStoreFileException("missing currentCycle, file=" + currentCycleFile);
             }
 
             switch (direction) {
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -500,6 +500,15 @@ class StoreTailer extends AbstractCloseable
     }
 
     private long nextIndexWithNextAvailableCycle(final int cycle) {
+        try {
+            return nextIndexWithNextAvailableCycle0(cycle);
+        } catch (MissingStoreFileException e) {
+            queue.refreshDirectoryListing();
+            return nextIndexWithNextAvailableCycle0(cycle);
+        }
+    }
+
+    private long nextIndexWithNextAvailableCycle0(final int cycle) {
         assert cycle != Integer.MIN_VALUE : "cycle == Integer.MIN_VALUE";
 
         if (cycle > queue.lastCycle() || direction == TailerDirection.NONE) {
--- a/src/test/java/net/openhft/chronicle/queue/RollCycleDefaultingTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/RollCycleDefaultingTest.java
@@ -4,6 +4,7 @@ import net.openhft.chronicle.core.time.TimeProvider;
 import net.openhft.chronicle.core.util.ObjectUtils;
 import net.openhft.chronicle.queue.impl.single.SingleChronicleQueueBuilder;
 import org.jetbrains.annotations.NotNull;
+import org.junit.After;
 import org.junit.Test;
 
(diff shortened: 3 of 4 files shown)


src/test/java/net/openhft/chronicle/queue/RollCycleDefaultingTest.java
```diff
@@ -15,6 +16,11 @@ public void alias() {
         assertEquals(RollCycles.class, ObjectUtils.implementationToUse(RollCycle.class));
     }
 
+    @After
+    public void clearDefaultRollCycleProperty() {
+        System.clearProperty(QueueSystemProperties.DEFAULT_ROLL_CYCLE_PROPERTY);
+    }
```
nicktindall, line 22: This was interfering with my test originally, I've since set the roll cycle explicitly, but felt like the right thing to do to clear it at the end.

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -189,6 +191,49 @@ public void firstAndLastIndexAreRefreshedAfterForceRefreshInterval() throws IOEx
...
+        tailingThroughDeletedCyclesWillRefreshThenRetry(qwcd -> qwcd.queue);
+    }
+
+    @Test
+    public void tailingThroughDeletedCyclesWillRefreshThenRetry_ReadOnly() throws IOException {
+        tailingThroughDeletedCyclesWillRefreshThenRetry(qwcd -> SingleChronicleQueueBuilder.binary(qwcd.queue.fileAbsolutePath())
+                .rollCycle(RollCycles.FAST_DAILY)
+                .readOnly(true)
+                .build());
+    }
```
nicktindall, line 205: This manifested in different ways for read-only and writable queues, I'm not entirely sure why, but I think it was to do with the fact they use different `DirectoryListing` implementations

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -500,6 +500,15 @@ private void inACycleFound(@NotNull final Bytes<?> bytes) {
     }
 
     private long nextIndexWithNextAvailableCycle(final int cycle) {
+        try {
+            return nextIndexWithNextAvailableCycle0(cycle);
+        } catch (MissingStoreFileException e) {
+            queue.refreshDirectoryListing();
+            return nextIndexWithNextAvailableCycle0(cycle);
+        }
+    }
```
nicktindall, line 509: This should only happen in scenarios where we used to throw an `IllegalStateException` or an `AssertionError` so it shouldn't affect the happy case, unless it impacts JIT in some mysterious way.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1223,7 +1223,7 @@ public int nextCycle(int currentCycle, @NotNull TailerDirection direction) {
 
                 if (!fileFound) {
                     directoryListing.refresh(true);
-                    throw new IllegalStateException(
+                    throw new MissingStoreFileException(
```
JerryShea, line 1226: nice

JerryShea (approved):
LGTM!

alamar (approved):

glukos (approved):

Issue #954, opened 2021-11-27
Add support for listening to messages as they are appended


peter-lawrey:
This functionality is useful for adding a listener to maintain a index of entries appended

Sets an AppenderListener which is called when an Excerpt is actually written.
This is called while the writeLock is still held, after the messages has been written.
For asynchronous writes, this is called in the background thread.

JerryShea:
Closing as I believe this has been merged. @peter-lawrey please re-open if not

hft-team-city:
Released in [Chronicle-Queue-5.23ea2](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.23ea2), [BOM-2.23ea25](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea25)

PR #955, opened 2021-11-27
Add support for listening to messages as they are appended #954

peter-lawrey:


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -164,13 +165,14 @@ protected SingleChronicleQueue(@NotNull final SingleChronicleQueueBuilder builde
             storeFactory = builder.storeFactory();
             checkInterrupts = builder.checkInterrupts();
             metaStore = builder.metaStore();
-            doubleBuffer = false; //builder.doubleBuffer();
+            doubleBuffer = builder.doubleBuffer();
```
nicktindall, line 172: Didn't this have a race condition in it?

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
```diff
@@ -3244,6 +3277,71 @@ protected SingleChronicleQueueBuilder builder(@NotNull File file, @NotNull WireT
...
+                    appenderListenerDump.append("idx: ").append(Long.toHexString(idx)).append("\n");
+                    w.bytes().readSkip(-4);
+                    appenderListenerDump.append(Wires.fromSizePrefixedBlobs(w)).append("\n");
+                })
+                .testBlockSize();
+    }
+
     @NotNull
     protected SingleChronicleQueueBuilder binary(@NotNull File file) {
         return builder(file, WireType.BINARY_LIGHT);
```
nicktindall, line 3356: This diff is impossible to review... is there a way we could do it with less change?

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -630,7 +630,9 @@ private void writeBytesInternal(@NotNull final BytesStore bytes, boolean metadat
             openContext(metadata, safeLength);
 
             try {
-                writeContext.wire().bytes().write(bytes);
+                final Bytes<?> bytes0 = writeContext.wire().bytes();
+                bytes0.readPosition(bytes0.writePosition());
```
nicktindall, line 639: Why do we set the `readPosition` here?

.github/workflows/build.yml
```diff
@@ -34,4 +34,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=OpenHFT_Chronicle-Queue
+        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=OpenHFT_Chronicle-Queue -Psonar -Pjava11
```
JerryShea, line 37: why -Pjava11?

Issue #949, opened 2021-11-15
The metadata file remains open despite closing a Queue


minborg:
Reproduce:

```
    @Test
    public void openFileTest() {
        File tmpDir = new File("open_file_test");

        try (final ChronicleQueue queue =
                     builder(tmpDir, wireType)
                             .build()) {

        }

        Jvm.pause(30_000);


    }
```
and performing:

```
% lsof | grep "metadata.cq4"


java      13682 pemi  txt       REG                1,5      131072            62645501 /Users/pemi/git/chronicle/opensource/Chronicle-Queue/open_file_test/metadata.cq4t

```

nicktindall:
Introduced in ce9885d65b0a7ba2921a32f5bc304298c323c06b

The change in `SingleTableBuilder` from 
`bytes = MappedBytes.mappedBytes(file, OS.SAFE_PAGE_SIZE, OS.SAFE_PAGE_SIZE, readOnly);`
to
`bytes = MappedBytes.singleMappedBytes(file, OS.SAFE_PAGE_SIZE * 2, readOnly);`
seems to be the culprit

PR #963, opened 2021-12-02
Test to show all the files can be deleted, would fail on windows otherwise.

peter-lawrey:
Issue #949


JerryShea (approved):

nicktindall:
Should we raise a separate issue to track the underlying cause? I thought this was due to SingleMappedBytes not closing files?

Issue #964, opened 2021-12-03
Allow named tailers to be used concurrently to split work


peter-lawrey:

RobAustin:
I'm going to close this, please reopen if not complete

JerryShea:
The PR is still open #965

minborg:
This is interesting and would be very useful for parallel `java.util.stream.Stream` wrappers.

tgd:
Closing this out as no work is planned on this issue at the moment. Please comment on this issue if you would like it re-opened.

PR #965, opened 2021-12-03
Allow named tailers to be used concurrently to split work #964

peter-lawrey:
Previously the behaviour of concurrent named tailers with the same name was undefined and not supported.
This change adds support for splitting work by named tailer by checking whether the index has changed during the call to readingDocument 
(changing the index in the table store at the end of readingDocument instead of close() )
I have added a test to show it detects and retries if the index is change while readingDocument is being performed.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -37,6 +37,7 @@ import static net.openhft.chronicle.wire.Wires.isEndOfFile;
  */
 class StoreTailer extends AbstractCloseable
         implements ExcerptTailer, SourceContext, ExcerptContext {
+    public static final long INVALID_INDEX = Long.MIN_VALUE;
     static final int INDEXING_LINEAR_SCAN_THRESHOLD = 70;
     static final StringBuilderPool SBP = new StringBuilderPool();
     static final EOFException EOF_EXCEPTION = new EOFException();
@@ -57,7 +58,7 @@ class StoreTailer extends AbstractCloseable
     private boolean readAfterReplicaAcknowledged;
     @NotNull
     private TailerState state = UNINITIALISED;
-    private long indexAtCreation = Long.MIN_VALUE;
+    private long indexAtCreation = INVALID_INDEX;
     private boolean readingDocumentFound = false;
     private long address = NO_PAGE;
     private boolean striding = false;
@@ -177,7 +178,7 @@ class StoreTailer extends AbstractCloseable
     @NotNull
     @Override
     public String toString() {
-        final long index = index();
+        final long index = this.index; // don't use index() as this confuses the debugger
         return "StoreTailer{" +
                 "index sequence=" + queue.rollCycle().toSequenceNumber(index) +
                 ", index cycle=" + queue.rollCycle().toCycle(index) +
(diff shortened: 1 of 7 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -195,6 +201,25 @@ public DocumentContext readingDocument(final boolean includeMetaData) {
         return documentContext;
     }
 
+    DocumentContext readingDocumentNamed(final boolean includeMetaData) {
+        for (int i = 0; i < 100; i++) {
```
rogersimmons: Under what conditions would this loop run to completion and throw the AssertionError?
Stopping after 100 iterations looks arbitrary. Is there something more robust we can do here?
It feels that the loop itself should retry indefinitely, and either succeed eventually, or internally detect any genuine error condition and throw from the body

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -195,6 +201,25 @@ public DocumentContext readingDocument(final boolean includeMetaData) {
         return documentContext;
```
nicktindall, line 202: Do we need to replicate this check in the named version? or can we remove it from unnamed? it makes it more confusing if they're inconsistent

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -1083,10 +1121,11 @@ private boolean tryWindBack(final int cycle) {
     }
 
     void index0(final long index) {
-        if (indexValue == null)
+        if (indexValue == null) {
             this.index = index;
-        else
-            indexValue.setValue(index);
+        } else if (!indexValue.compareAndSwapValue(this.indexChecker, index)) {
+            this.indexChecker = Long.MIN_VALUE; // invalid.
```
nicktindall, line 1135: Would it be better to give this a constant with a meaningful name? It's a bit hard to follow what the significance of Long.MIN_VALUE is.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -195,6 +201,25 @@ public DocumentContext readingDocument(final boolean includeMetaData) {
...
+
+            if (indexChecker != Long.MIN_VALUE) {
+                this.index = indexChecker;
+                if (context.isPresent() && !context.isMetaData())
+                    incrementIndex();
+                return documentContext;
+            }
+            documentContext.close();
+        }
+        throw new AssertionError();
```
nicktindall, line 223: Is it possible in cases of high contention that you'd get to the end of the 100 iterations? Probably very unlikely but even if it's a remote possibility we should probably put a meaningful exception/message in here.

nicktindall:
Apologies for pushing to the branch, but I just added a test to the test class that runs concurrent competing consumer threads and it doesn't seem to work. Perhaps I've missed something, alarmingly the test breaks even for a single consumer. ~I will note there's another test that relies on named tailers which displays similar behaviour occasionally (in QE, `UngracefulShutdownTest`, the same excerpt is read twice). I had always assumed the test was broken but maybe named tailers is?~
I checked and this test passes on develop, so whatever the issue is, it's introduced (or made significantly more likely) by the changes on this branch

src/test/java/net/openhft/chronicle/queue/ConcurrentNamedTailersTest.java
```diff
@@ -0,0 +1,303 @@
...
+             final ExcerptAppender appender = q.acquireAppender()) {
+            for (int i = 0; i < numberOfEntries; i++) {
+                try (final DocumentContext documentContext = appender.writingDocument()) {
+                    documentContext.wire().write("index").int64(i);
+                }
+            }
+        }
+        Jvm.startup().on(ConcurrentNamedTailersTest.class, "Wrote " + numberOfEntries + " entries");
+
+        final int numReaders = 6;
```
JerryShea, line 162: this fails on my machine when numReaders = 1

RobAustin (approved):

tgd:
Closing PR in order to close issue - we can re-open as needed

PR #966, opened 2021-12-03
Write less messages in case of timeout on github build server

peter-lawrey:


peter-lawrey:
Replaced by a later change

Issue #975, opened 2021-12-15
SingleChronicleQueue#exceptsPerCycle returning obsolete value


alamar:
When doing a lot of background writing to Queue, the `queue.exceptsPerCycle(cycle)` invocation may return a value that is smaller by `1` than the last actual written entry index.

The underlying problem is that `tailer.store.lastSequenceNumber(tailer)` returns incorrect value in its memory lookup loop, whereas the fallback `tailer.store.sequenceForPosition(tailer, Long.MAX_VALUE, false)` will return correct value.

The method name is semantically problematic (it does not return value 'per' cycle, excerpts are pull-based so counting them is confusing, and there's a typo) and it is the last user of obsolete `acquireTailer()` so maybe we should rethink this method over.

hft-team-city:
Released in [Chronicle-Queue-5.22ea16](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.22ea16), [BOM-2.22ea106](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea106)

PR #976, opened 2021-12-15
Test for #975

alamar:
#975

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/ExcerptsPerCycleTestOurs.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/ExcerptsPerCycleTestOurs.java
@@ -1,0 +1,114 @@
+package net.openhft.chronicle.queue.impl.single;
+import net.openhft.chronicle.core.io.IOTools;
+import net.openhft.chronicle.queue.ExcerptTailer;
+import net.openhft.chronicle.queue.RollCycle;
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/queue/impl/single/ExcerptsPerCycleTestOurs.java
```diff
@@ -0,0 +1,114 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.StreamCorruptedException;
+import java.nio.file.Files;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.LongConsumer;
+
+public final class ExcerptsPerCycleTestOurs {
```
nicktindall, line 19: I don't think this name will get picked up by surefire?

alamar: I'm not sure it will be merged under this name, but new JUnit is annotations-based and it is present.

nicktindall: I was referring to surefire inclusion, does that change when you're using Jupiter somehow?
https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

PR #977, opened 2021-12-16
add performance log for normaliseEOFs and javadoc

JerryShea:
We had a problem at a customers where they were creating a lot of appenders and `normaliseEOFs` was taking a long time to complete - so long, in fact, that other threads' WriteLocks were timing out. 

This PR contains a warning for slow `normaliseEOF`.


nicktindall (approved):

rogersimmons (approved):
Have approved this particular change, but I'm keen to understand why normaliseEOFs is taking excessively long in some cases. Do the users have unusually large cycle sets, or are the indexes unusually sparse so tracking to the end is taking a long time, or is something else stalling this? Conceptually the normaliseEOF step is straightforward so excessive runtime suggests something else in the environment may not be appropriately configured.

JerryShea:
@rogersimmons yes so am I! I'll give you some more background by email

PR #983, opened 2021-12-23
Exact exerpts count as well as optimized approximate

### LINKED ISSUE #975: SingleChronicleQueue#exceptsPerCycle returning obsolete value

alamar:
closes #975

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
@@ -658,12 +658,12 @@ class SCQIndexing extends AbstractCloseable implements Demarshallable, WriteMars
         return (index & (indexSpacing - 1)) == 0;
     }
 
-    public long lastSequenceNumber(@NotNull ExcerptContext ec)
+    public long lastSequenceNumber(@NotNull ExcerptContext ec, boolean approximate)
             throws StreamCorruptedException {
         throwExceptionIfClosed();
 
         Sequence sequence1 = this.sequence;
-        if (sequence1 != null) {
+        if (approximate && sequence1 != null) {
             for (int i = 0; i < 128; i++) {
 
                 long address = writePosition.getVolatileValue(0);
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -572,16 +572,42 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
         return pool.nextCycle(cycle, direction);
     }
 
+    /**
+     * Instead, {@link #approximateExcerptsInCycle(int)} or {@link #exactExcerptsInCycle(int)} should be called.
+     */
+    @Deprecated(/* For removal in x.23 */)
     public long exceptsPerCycle(int cycle) {
+        return approximateExcerptsInCycle(cycle);
+    }
+
+    /**
+     * Returns a number of excerpts in a cycle. May use a fast path to return the cycle length cached in indexing,
+     * which is updated last during append operation so may be possible that a single entry is available for reading
+     * but not acknowledged by this method yet.
+     */
+    public long approximateExcerptsInCycle(int cycle) {
         throwExceptionIfClosed();
         // TODO: this function may require some re-work now that acquireTailer has been deprecated
         StoreTailer tailer = acquireTailer();
         try {
-            if (tailer.moveToCycle(cycle)) {
-                return tailer.store.lastSequenceNumber(tailer) + 1;
-            } else {
-                return -1;
-            }
+            return tailer.moveToCycle(cycle) ? tailer.store.approximateLastSequenceNumber(tailer) + 1 : -1;
+        } catch (StreamCorruptedException e) {
+            throw new IllegalStateException(e);
+        } finally {
+            tailer.releaseStore();
+        }
+    }
+
+    /**
+     * Returns an exact number of excerpts in a cycle available for reading. This may be a computationally
+     * expensive operation.
+     */
+    public long exactExcerptsInCycle(int cycle) {
+        throwExceptionIfClosed();
+        // TODO: this function may require some re-work now that acquireTailer has been deprecated
+        StoreTailer tailer = acquireTailer();
+        try {
+            return tailer.moveToCycle(cycle) ? tailer.store.exactLastSequenceNumber(tailer) + 1 : -1;
         } catch (StreamCorruptedException e) {
             throw new IllegalStateException(e);
         } finally {
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
@@ -304,10 +304,19 @@ public class SingleChronicleQueueStore extends AbstractCloseable implements Wire
     }
 
     @Override
+    @Deprecated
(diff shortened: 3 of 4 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -658,12 +658,12 @@ public boolean indexable(long index) {
         return (index & (indexSpacing - 1)) == 0;
     }
 
-    public long lastSequenceNumber(@NotNull ExcerptContext ec)
+    public long lastSequenceNumber(@NotNull ExcerptContext ec, boolean approximate)
```
RobAustin, line 661: please add javadoc to this method, I guess stating that approximate is faster

RobAustin, line 661: sorry - ignore my comment about java doc as its an impl method

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -572,13 +572,47 @@ public int nextCycle(int cycle, @NotNull TailerDirection direction) throws Parse
...
+     * Returns a number of excerpts in a cycle. May use a fast path to return the cycle length cached in indexing,
+     * which is updated last during append operation so may be possible that a single entry is available for reading
+     * but not acknowledged by this method yet.
+     */
+    public long approximateExcerptsInCycle(int cycle) {
+        throwExceptionIfClosed();
+        // TODO: this function may require some re-work now that acquireTailer has been deprecated
+        StoreTailer tailer = acquireTailer();
+        try {
+            if (tailer.moveToCycle(cycle)) {
```
RobAustin: consider -> return tailer.moveToCycle(cycle) ? tailer.store.approximateLastSequenceNumber(tailer) + 1 : -1;

RobAustin: which should be the same thing

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -572,13 +572,47 @@ public int nextCycle(int cycle, @NotNull TailerDirection direction) throws Parse
...
+    /**
+     * Returns an exact number of excerpts in a cycle available for reading. This may be a computationally
+     * expensive operation.
+     */
+    public long exactExcerptsInCycle(int cycle) {
         throwExceptionIfClosed();
         // TODO: this function may require some re-work now that acquireTailer has been deprecated
         StoreTailer tailer = acquireTailer();
         try {
             if (tailer.moveToCycle(cycle)) {
```
RobAustin: likewise here -> return <> ? <> : <>

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
```diff
@@ -304,10 +304,21 @@ public long sequenceForPosition(@NotNull final ExcerptContext ec, final long pos
...
 
     @Override
+    @Deprecated
     public long lastSequenceNumber(@NotNull ExcerptContext ec) throws StreamCorruptedException {
+        return approximateLastSequenceNumber(ec);
+    }
+
+    public long approximateLastSequenceNumber(@NotNull ExcerptContext ec) throws StreamCorruptedException {
+        throwExceptionIfClosedInSetter();
+
```
RobAustin: remove white space

src/test/java/net/openhft/chronicle/queue/impl/single/EntryCountNotBehindReadTest.java
```diff
@@ -0,0 +1,135 @@
...
+                        onRead.accept(entry.index());
+                        ++count;
+                    }
+                }
+            }
+        }
+    }
+
+    private static void waitOn(CyclicBarrier barrier) {
+
```
RobAustin: remove whitespace

src/test/java/net/openhft/chronicle/queue/impl/single/EntryCountNotBehindReadTest.java
```diff
@@ -0,0 +1,135 @@
...
+        new Thread(() -> {
+            waitOn(startBarrier);
+            for (int i = 0; i < TOTAL_EVENTS; ++i) {
+                queue.acquireAppender().writingDocument().close();
+            }
+        }).start();
+    }
+
+    private void runReader(SingleChronicleQueue queue, CyclicBarrier startBarrier, LongConsumer onRead) {
+
```
RobAustin: remove whitespace

src/test/java/net/openhft/chronicle/queue/impl/single/EntryCountNotBehindReadTest.java
```diff
@@ -0,0 +1,135 @@
...
+        try (ExcerptTailer tailer = queue.createTailer()) {
+            if (tailer.moveToCycle(cycle)) {
+                excerptCount = cycleType.toSequenceNumber(tailer.toEnd().index());
+            }
+        }
+        assertFalse(readCount > excerptCount);
+    }
+
+    private void startWriter(SingleChronicleQueue queue, CyclicBarrier startBarrier) {
+
```
RobAustin: remove whitespace

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -572,16 +572,42 @@ public int nextCycle(int cycle, @NotNull TailerDirection direction) throws Parse
         return pool.nextCycle(cycle, direction);
     }
 
+    @Deprecated
```
minborg: We should indicate "for removal" and when the method is to be removed, for example "in x.23"

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
minborg, line 661: We should provide a compatible `lastSequenceNumber(@NotNull ExcerptContext ec)` so that old code using the signature of the old method can compile. This method can then call the new method with an appropriate value of the `approximate` parameter.

minborg: We should invoke using `approximate = true` to retain the old behaviour.

minborg (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
alamar: It is a non-public class, I don't think that we need to care about code calling it, unless it's our code.

alamar: It's package private and the method is not part of any interface.

minborg: Great. That's perfect. We do not have to do anything.

PR #987, opened 2021-12-31
Feature/only unlock if process dead

JerryShea:


RobAustin (changes requested):
we discussed this in the dev meeting, and the consensus was that we should consider other approaches than throwing an UnrecoverableTimeoutException() and forcing the user to run some other code ( such as Unlock Main ) to remove the lock.

JerryShea:
Let's discuss at dev meeting again this week. Only behaviour change introduced in this PR is if the new `queue.lock.recover.dead.process` parameter is set, and then we behave consistently with the append lock i.e. we unlock if the locking process is dead, otherwise throw exception

RobAustin (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -88,11 +88,16 @@ public void lock() {
...
-            tlPauser.reset();
-            lock();
+            else if (onlyUnlockIfProcessDead) {
+                if (forceUnlockIfProcessIsDead())
+                    lock();
+                else
+                    throw new UnrecoverableTimeoutException(new IllegalStateException(warningMsg + UNLOCK_MAIN_MSG));
+            } else {
+                warn().on(getClass(), warningMsg + UNLOCKING_FORCIBLY_MSG);
+                forceUnlock(value);
```
nicktindall: I feel like the value variable could be named better (e.g. currentLockValue or currentPID or whatever it is)

nicktindall: Especially now we're increasing the number of lines between where it's assigned and where it's used

nicktindall (approved):
I like this change. Perhaps it should become default in .23?

Although if we do that maybe we should have config like `forceUnlockOnTimeoutWhen` = `[ALWAYS | NEVER | LOCKING_PROCESS_DEAD]` rather than having `queue.lock.recover.dead.process` and `queue.dont.recover.lock.timeout` as separate and possibly inconsistent confgs

JerryShea:
> I like this change. Perhaps it should become default in .23?
> 
> Although if we do that maybe we should have config like `forceUnlockOnTimeoutWhen` = `[ALWAYS | NEVER | LOCKING_PROCESS_DEAD]` rather than having `queue.lock.recover.dead.process` and `queue.dont.recover.lock.timeout` as separate and possibly inconsistent confgs

We normally give users two releases before we change things like this, but happy to take votes from the group.

I have implemented the enum which was very helpful - it helped shake out some inconsistencies (and further reinforces the case to refactor these locking classes)

src/main/java/net/openhft/chronicle/queue/impl/single/TSQueueLock.java
```diff
@@ -122,15 +121,25 @@ public void waitForLock() {
         }
     }
 
+    private void warnAndForceUnlock(String msg, long value) {
+        warnLock(msg, value);
```
JerryShea, line 125: duplicated but not quite the same code - we should address with https://github.com/OpenHFT/Chronicle-Queue/issues/994

src/test/java/net/openhft/chronicle/queue/impl/single/TSQueueLockTest.java
```diff
@@ -0,0 +1,284 @@
...
+import java.util.Collection;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.*;
+
+public class TSQueueLockTest extends QueueTestCommon {
```
JerryShea, line 32: this is copy/pasted from TableStoreWriteLockTest. This test definitely needs some more work/refactoring - see https://github.com/OpenHFT/Chronicle-Queue/issues/994

PR #989, opened 2022-01-13
Add read write buffer mode params to benchmark

nicktindall:
- Allow specification of readBufferMode and writeBufferMode in `QueueMultiThreadedJLBHBenchmark`
- Fix bug when determining whether document was read
- Allow specification of test class for output

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/bench/QueueMultiThreadedJLBHBenchmark.java
+++ b/src/test/java/net/openhft/chronicle/queue/bench/QueueMultiThreadedJLBHBenchmark.java
@@ -28,10 +28,7 @@ import net.openhft.chronicle.jlbh.JLBH;
 import net.openhft.chronicle.jlbh.JLBHOptions;
 import net.openhft.chronicle.jlbh.JLBHTask;
 import net.openhft.chronicle.jlbh.TeamCityHelper;
-import net.openhft.chronicle.queue.ExcerptAppender;
-import net.openhft.chronicle.queue.ExcerptTailer;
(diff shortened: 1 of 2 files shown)


src/test/java/net/openhft/chronicle/queue/bench/QueueMultiThreadedJLBHBenchmark.java
```diff
@@ -111,7 +115,7 @@ public void init(JLBH jlbh) {
                 while (!stopped) {
                     long beforeReadNs = System.nanoTime();
                     try (DocumentContext dc = tailer.readingDocument()) {
-                        if (dc.wire() == null)
+                        if (!dc.isPresent())
```
nicktindall, line 118: `dc.wire() == null` is not a valid test when using the ring buffer

JerryShea: nice

JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/bench/QueueMultiThreadedJLBHBenchmark.java
minborg: I think the `wire == null` pattern is used all over the place. Should we update all of them (code, demos, docs etc) to `isPresent()`?

JerryShea: +1 @minborg nice idea

PR #990, opened 2022-01-13
Uncomment test, add expected warning

### LINKED ISSUE #326: Issues with the pretoucher when used for reading instead of writing.
The problem in the second one occurs as the underlying MappedBytes is changing so ti has to reset the position for the new MappedBytes object.  It appears the file is the same so it's not clear why it would create a new one. Continuing to investigate.

alamar:
fixes #326

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/ValueStringArrayTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/ValueStringArrayTest.java
@@ -13,9 +13,10 @@ public class ValueStringArrayTest extends ChronicleQueueTestBase {
     private static final String EXPECTED = "hello world";
     private final ValueStringArray using = new ValueStringArray();
 
-    @Ignore("see https://github.com/OpenHFT/Chronicle-Bytes/issues/326")
     @Test
     public void test() {
+        // No explicit support of putting a Value into Wire.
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/queue/ValueStringArrayTest.java
```diff
@@ -13,9 +13,11 @@
     private static final String EXPECTED = "hello world";
     private final ValueStringArray using = new ValueStringArray();
 
-    @Ignore("see https://github.com/OpenHFT/Chronicle-Bytes/issues/326")
     @Test
     public void test() {
+        // No explicit support of putting a Value into Wire.
+        expectException("BytesMarshallable found in field which is not matching exactly, the object may not " +
```
minborg: Can we make the text shorter as any change in the output would make this expectation fail again?

minborg (approved):

src/test/java/net/openhft/chronicle/queue/ValueStringArrayTest.java
alamar: OK, since this is a different module, I will shorten this message to avoid cascading failure.

PR #992, opened 2022-01-17
warn if cleanupStoreFilesWithNoData is slow

JerryShea:
Similar to https://github.com/OpenHFT/Chronicle-Queue/pull/977


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -925,6 +926,9 @@ void cleanupStoreFilesWithNoData() {
             writeLock.unlock();
             if(fireOnReleasedEvent != null)
                 BackgroundResourceReleaser.run(fireOnReleasedEvent);
+            long tookMillis = (System.nanoTime() - start) / 1_000_000;
+            if (tookMillis > 100)
```
rogersimmons: Default to 100, but I would suggest this can be overridden through a property or similar

rogersimmons (approved):

nicktindall (approved):

PR #997, opened 2022-01-26
Remove exception suppressing clause

### LINKED ISSUE #996: Remove exception-suppressing code from SCQIndexing#moveToIndex0
```
        } catch (IllegalStateException e) {
            return linearScan(ec.wire(), index, -1, 0);
        }
````
It was introduced in #334 to support lazy indexing, which was in turn removed in 5.x
Removing this clause does not seem to affect any tests in Queue and dependent projects, but we have observed in thread dumps and have suspicion that it may be masking problems (such as accessing closed queue) and make Queue hang instead of throwing an exception.

alamar:
fixes #996


peter-lawrey:
This fallback applies if no indexed entries are found. Instead, it does a brute force search. Perhaps this obscures an underlying problem.

alamar:
@peter-lawrey do you approve? The idea is to un-obscure the problem which we otherwise cannot see.

JerryShea (approved):
Totally agree with the sentiment although not really close enough to this code to give more than a qualified approval!

Issue #998, opened 2022-02-01
IllegalStateException: expecting index, was header in CycleNotFoundTest


alamar:
#996 has unmasked the following exception which will very rarely occur in `net.openhft.chronicle.queue.CycleNotFoundTest#tailerCycleNotFoundTest` (on TC and in IDEA)

```
 Caused by: java.lang.IllegalStateException: expecting index, was header, bytes: 
  00000000             b9 06 68 65  61 64 65 72 b6 08 53 43     ··he ader··SC
  00000010 51 53 74 6f 72 65 82 56  00 00 00 8f 8f 8f 8f 8d QStore·V ········
  00000020 02 00 00 00 00 00 00 00  02 00 00 00 00 00 00 00 ········ ········
  00000030 00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00 ········ ········
  00000040 b6 0c 53 43 51 53 49 6e  64 65 78 69 6e 67 82 1d ··SCQSIn dexing··
  00000050 00 00 00 a2 00 80 04 a7  00 00 00 00 00 00 00 00 ········ ········
  00000060 8e 02 00 00 00 00 00 a7  00 00 00 00 00 00 00 00 ········ ········
  00000070 01                                               ·
    at net.openhft.chronicle.queue.impl.single.SCQIndexing.array(SCQIndexing.java:199)
    at net.openhft.chronicle.queue.impl.single.SCQIndexing.getIndex2index(SCQIndexing.java:579)
    at net.openhft.chronicle.queue.impl.single.SCQIndexing.moveToIndex0(SCQIndexing.java:278)
    at net.openhft.chronicle.queue.impl.single.SCQIndexing.moveToIndex(SCQIndexing.java:254)
    at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueStore.moveToIndexForRead(SingleChronicleQueueStore.java:228)
    at net.openhft.chronicle.queue.impl.single.StoreTailer.moveToIndexResult0(StoreTailer.java:668)
    at net.openhft.chronicle.queue.impl.single.StoreTailer.moveToIndexInternal(StoreTailer.java:739)
    at net.openhft.chronicle.queue.impl.single.StoreTailer.next0(StoreTailer.java:294)
    at net.openhft.chronicle.queue.impl.single.StoreTailer.readingDocument0(StoreTailer.java:215)
[...]

PR #999, opened 2022-02-01
 After segment has just been rolled over, index is not initialized yet

alamar:
Test which will catch #998 much more eagerly, once per 10-20 runs

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
@@ -273,6 +273,8 @@ class SCQIndexing extends AbstractCloseable implements Demarshallable, WriteMars
     // visible for testing
     @Nullable
     ScanResult moveToIndex0(@NotNull final ExcerptContext ec, final long index) {
+        if (index2Index.getVolatileValue() == NOT_INITIALIZED)
+            return null;
 
         Wire wire = ec.wireForIndex();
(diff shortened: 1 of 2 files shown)


JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -273,6 +273,8 @@ private ScanResult moveToIndexFromTheStart(@NotNull ExcerptContext ec, long inde
     // visible for testing
     @Nullable
     ScanResult moveToIndex0(@NotNull final ExcerptContext ec, final long index) {
+        if (index2Index.getVolatileValue() == NOT_INITIALIZED)
```
JerryShea, line 276: fantastic, although again I am not as close to this code as I would like, so take my approval with a grain of sand

JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/CycleNotFoundTest.java
```diff
@@ -52,7 +52,7 @@ public void tailerCycleNotFoundTest() throws InterruptedException, ExecutionExce
         Runnable reader = () -> {
```
glukos, line 52: Looks like there's a typo in Executors.newFixedThreadPool call - we pass `NUMBER_OF_MSG` instead of `NUMBER_OF_TAILERS`

PR #1006, opened 2022-02-14
Feature/issue1005 attempt deflake

JerryShea:
Three changes in this PR:
1. some tests were calling AbstractCloseable.enableCloseableTracing() and AbstractCloseable.assertCloseablesClosed(); when they didn't need to (the base test class does)
2. remove the unused `writeOneThenWait` property from `RollCycleMultiThreadStressTest` and add back in diagnostic `Jvm.perf` output in case flakiness returns
3. reduce the number of cores that `RollCycleMultiThreadStressTest` tries to use. It tries to use _all_ the cores on the host and on a docker host Runtime.getRuntime().availableProcessors() returns the parent's cores. I have reduced this with `system.properties` and these tests run much more reliably now
I have not done much research into this but I believe Runtime.getRuntime().availableProcessors() behaves better on docker after Java 10. We may need to consider some changes in this area.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleTest.java
@@ -1,13 +1,11 @@
 package net.openhft.chronicle.queue.impl.single;
 
-import net.openhft.chronicle.core.io.AbstractCloseable;
 import net.openhft.chronicle.core.time.SetTimeProvider;
 import net.openhft.chronicle.core.time.TimeProvider;
 import net.openhft.chronicle.queue.*;
 import net.openhft.chronicle.queue.impl.StoreFileListener;
 import org.jetbrains.annotations.NotNull;
 import org.junit.After;
-import org.junit.Before;
 import org.junit.Test;
 
 import java.io.File;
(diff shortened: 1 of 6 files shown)


nicktindall:
The docker agent has a property `teamcity.agent.hardware.cpuCount`

Could we pass that in via the environment and use that to set available CPUs? @alamar is docker restricting the CPUs usable by the agent and is that property the number of cores it's limited to?

RobAustin:
> The docker agent has a property `teamcity.agent.hardware.cpuCount`
> 
> Could we pass that in via the environment and use that to set available CPUs? @alamar is docker restricting the CPUs usable by the agent and is that property the number of cores it's limited to?

see  https://stackoverflow.com/questions/37871540/how-many-cpus-does-a-docker-container-use

(A docker container is just a bunch of private namespaces -- since it isn't emulating CPUs, it would need to go out of its way to impose any restrictions on them; Linux kernels provide facilities to allow such going-out-of-one's-way, but that's still something that needs to actually be explicitly performed).

RobAustin (approved):

nicktindall:
> > The docker agent has a property `teamcity.agent.hardware.cpuCount`
> > 
> > Could we pass that in via the environment and use that to set available CPUs? @alamar is docker restricting the CPUs usable by the agent and is that property the number of cores it's limited to?
> 
> see  https://stackoverflow.com/questions/37871540/how-many-cpus-does-a-docker-container-use
> 
> (A docker container is just a bunch of private namespaces -- since it isn't emulating CPUs, it would need to go out of its way to impose any restrictions on them; Linux kernels provide facilities to allow such going-out-of-one's-way, but that's still something that needs to actually be explicitly performed). 

Yes but docker has flags that allow you to limit the resources available to the container. I had no involvement in setting up the docker agents but I assume this manifests as an option in the agent configuration? Just wondering if we set that and, if so, how to communicate that with Java (that doesn't see this stuff till 10)
See https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler

PR #1010, opened 2022-02-18
Adjust test for non-lenient readByte()

alamar:
see OpenHFT/Chronicle-Bytes#329


src/test/java/net/openhft/chronicle/queue/WriteBytesTest.java
```diff
@@ -272,25 +272,25 @@ public void testWriteBytesAndDump() {
                     "# # EndOfFile\n" +
                     "# position: 992, header: 18\n" +
                     "--- !!data #binary\n" +
-                    "-753555055760.82\n" +
+                    "# # EndOfFile\n" +
```
minborg, line 275: Not sure why this now reads EndOfFile ?

alamar: Because we have introduced non-lenient readByte() mode (the default) which is used in parseDouble(). These malformed records seem to return EOF.

minborg (approved):

PR #1012, opened 2022-02-24
Remove deprecated methods and classes in 2.23

alamar:


src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -264,12 +264,12 @@ public ChronicleReader historyRecords(final long maxHistoryRecords) {
     }
 
     public ChronicleReader asMethodReader(@Nullable String methodReaderInterface) {
-        if (methodReaderInterface == null)
-            entryHandlerFactory = () -> QueueEntryHandler.dummy(wireType);
```
JerryShea, line 268: what if it is null?

JerryShea, line 268: what if it is null?

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
alamar: Default value of the field is retained

PR #1032, opened 2022-03-31
ChronicleReader content-based limiter

nicktindall:
I made an attempt at tidying up the ChronicleReader a bit as part of this PR. It's a tricky task as it's evolved somewhat "organically". I think I've preserved all functionality and reduced the complexity of that main loop a little, but there's definitely more work to be done.


src/main/java/net/openhft/chronicle/queue/internal/reader/queueentryreaders/MethodReaderQueueEntryReader.java
```diff
@@ -0,0 +1,52 @@
...
+import net.openhft.chronicle.wire.WireType;
+
+import java.nio.ByteBuffer;
+
+public final class MethodReaderQueueEntryReader implements QueueEntryReader {
+
+    private final ExcerptTailer tailer;
+    private final MessageConsumer messageConsumer;
+    private final MethodReader methodReader;
+    private final Bytes<ByteBuffer> bytes;
```
nicktindall, line 21: Should probably clean this up. Perhaps make the thing Closeable.

src/main/java/net/openhft/chronicle/queue/internal/reader/queueentryreaders/VanillaQueueEntryReader.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private int runMessageConverter(DocumentContext documentContext) {
+        AtomicInteger matchCounter = tlMatchCounter.get();
+        matchCounter.set(0);
+        messageConverter.accept(documentContext.wire(), val -> {
+            if (messageConsumer.consume(documentContext.index(), val)) {
+                matchCounter.incrementAndGet();
+            }
+        });
```
nicktindall: This is a bit boiler-platey due to the mismatch between what the `messageConverter` expects to write to and the new MessageConsumer interface.

src/main/java/net/openhft/chronicle/queue/internal/reader/queueentryreaders/CustomPluginQueueEntryReader.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    protected ReadResult doRead(DocumentContext documentContext) {
+        AtomicInteger matchCounter = tlMatchCounter.get();
+        matchCounter.set(0);
+        plugin.onReadDocument(documentContext, value -> {
+            if (consumer.consume(documentContext.index(), value)) {
+                matchCounter.incrementAndGet();
+            }
+        });
+        return new ReadResult(true, matchCounter.get());
```
minborg: Creates an object.

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -191,23 +171,50 @@ public void execute() {
...
+                return new VanillaQueueEntryReader(tailer, pollMethod, entryHandlerFactory.get(), messageConsumer);
             } else {
-                customPlugin.onReadDocument(dc, messageConsumer);
+                return new CustomPluginQueueEntryReader(tailer, pollMethod, customPlugin, messageConsumer);
             }
+        } else {
+            return new MethodReaderQueueEntryReader(tailer, messageConsumer, wireType, methodReaderInterface, showMessageHistory);
         }
-        return true;
+    }
```
nicktindall, line 200: There are essentially 3 different ways of turning entries from the queue into strings, they're "vanilla", customPlugin and methodReader... I think having the separate `QueueEntryReaders` for each type is nice because it makes it clear which settings are relevant for each type. It's a bit ugly because there are subtle differences between the three approaches. Hopefully this structure makes it easier to identify functionality we don't need.

src/main/java/net/openhft/chronicle/queue/reader/ReadResult.java
```diff
@@ -0,0 +1,21 @@
...
+        this.matches = matches;
+    }
+
+    public boolean found() {
+        return found;
+    }
+
+    public int matches() {
+        return matches;
+    }
```
minborg: If we make this class mutable, we could hold TL instances to prevent creating new objects. That has other drawbacks of course.

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -492,15 +516,16 @@ private ChronicleQueue createQueue() {
                 .build();
     }
 
-    protected void applyFiltersAndLog(final String text, final long index, AtomicLong matches) {
+    private boolean applyFiltersAndLog(final long index, final String text) {
```
nicktindall: I changed this API so the MessageConsumer is no longer responsible for updating the match counter, instead a boolean is returned to indicate whether something made it to the sink or not. The caller keeps track of how many made it.

src/main/java/net/openhft/chronicle/queue/reader/ContentBasedLimiter.java
```diff
@@ -0,0 +1,10 @@
+package net.openhft.chronicle.queue.reader;
+
+import net.openhft.chronicle.wire.DocumentContext;
+
+public interface ContentBasedLimiter {
+
+    boolean shouldHaltReading(DocumentContext documentContext);
```
nicktindall, line 17: This is kind-of a predicate, but with a proper name, it's clearer what the return value means.

src/main/java/net/openhft/chronicle/queue/reader/MessageConsumer.java
```diff
@@ -0,0 +1,18 @@
+package net.openhft.chronicle.queue.reader;
+
+/**
+ * Message consumers make a chain-of-responsibility pattern, they
```
nicktindall, line 4: They don't yet, but they could - you could do includeFilter -> excludeFilter -> index Prepender -> sink

I just left it as it was to minimise the change

src/main/java/net/openhft/chronicle/queue/reader/QueueEntryReader.java
```diff
@@ -0,0 +1,6 @@
+package net.openhft.chronicle.queue.reader;
+
+public interface QueueEntryReader {
+
+    ReadResult read();
```
nicktindall: Returning a ReadResult means we can indicate if an entry was found, and how many outputs that entry produced then the main loop can keep track of limits and what to do when we reach the end of the queue.

nicktindall: Returning a ReadResult means we can indicate if an entry was found, and how many outputs that entry produced then the main loop can keep track of limits and what to do when we reach the end of the queue.

nicktindall: Returning a ReadResult means we can indicate if an entry was found, and how many outputs that entry produced then the main loop can keep track of limits and what to do when we reach the end of the queue.

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -123,45 +125,23 @@ public void execute() {
...
-                                return found;
-                            };
-                        }
 
                         while (!Thread.currentThread().isInterrupted()) {
-                            boolean found = readOne.getAsBoolean();
+                            if (shouldHaltReadingDueToContentBasedLimit(tailer)) {
+                                running = false;
+                                break;
+                            }
```
nicktindall, line 123: This isn't ideal because we read every message in the queue twice; once to determine whether we've reached the content based limit, then again to print the entry to the sink. 

The reason for this is the MethodReader never gives us access to the Wire, I could refactor it so we use Wire.methodReader(,,) with the readSkip(-4) trick instead of using a tailer.methodReader() and we might be able to combine the content-based limit and the conversion to a string against the same wire, resetting the positions inbetween.

nicktindall: I wonder if there's a less hacky way to do readDocument() on a DocumentContext wire than calling readSkip(-4), it kind of feels brittle or hacky.

src/main/java/net/openhft/chronicle/queue/internal/reader/queueentryreaders/CustomPluginQueueEntryReader.java
nicktindall: Do we need to care about garbage for ChronicleReader? isn't it a command line tool for dumping the contents of a queue?

minborg: That's right. Then this is perfectly ok.

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
JerryShea: nice - much simpler

pom.xml
```diff
@@ -308,7 +308,8 @@
                         <configuration>
                             <referenceVersion>5.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>100.0</binaryCompatibilityPercentageRequired>
+                            <!-- Reduce the API surface of ChronicleReader -->
+                            <binaryCompatibilityPercentageRequired>86.1</binaryCompatibilityPercentageRequired>
```
JerryShea, line 312: This is a significant decrease! @minborg will be happy!

minborg: It would be interesting to see what kind of API compatibility issues were detected. Could you paste them here in a comment?

src/main/java/net/openhft/chronicle/queue/internal/reader/queueentryreaders/CustomPluginQueueEntryReader.java
nicktindall: Managed to remove the object creation anyway :stuck_out_tongue_winking_eye:

src/main/java/net/openhft/chronicle/queue/reader/ReadResult.java
nicktindall: Got rid of it

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
nicktindall: Kinda changed it back, but in a nicer way

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
nicktindall: The more I think about it, the less I care about reading each entry twice. Performance here isn't critical, it's really not much different to reading the enclosed Wire twice anyway, it means we don't have to do any hacky stuff, and it gives us more flexibility in how we implement QueueEntryReaders

pom.xml
nicktindall: Screenshot from 2022-04-01 14-14-14

nicktindall: I could restore readOne, but I don't know why it was exposed. If it's public it means you can continue to call it beyond when execute() would have stopped calling it. It seems very internal to me, unless I've missed the point of this tool?

nicktindall: `applyFiltersAndLog` being protected looks like it might have been a way to override what happens to each message, we can do that better by adding a `MessageConsumer` as a parameter if it's a desired thing?

src/main/java/net/openhft/chronicle/queue/ChronicleReaderMain.java
```diff
@@ -177,6 +177,8 @@ protected Options options() {
         addOption(options, "k", "reverse", false, "Read the queue in reverse", false);
         addOption(options, "h", "help-message", false, "Print this help and exit", false);
         addOption(options, "x", "max-results", true, "Limit the number of results to output", false);
+        addOption(options, "cbl", "content-based-limiter", true, "Specify a content-based limiter", false);
+        addOption(options, "cblArg", "content-based-limiter-argument", true, "Specify an argument for use by the content-based limiter", false);
```
nicktindall, line 181: I think we need to move away from one letter args. The time has come!

pom.xml
JerryShea: I know customers (and our products) definitely descend from ChronicleReaderMain, but that has not changed.

I believe that customers also descend from ChronicleReader - it is possible they use readOne, but I don't think so. readOne was added in https://github.com/OpenHFT/Chronicle-Queue/commit/8e4d7a5a2d92c8dcc561c25f96ddaccc931b6867 by @peter-lawrey

JerryShea (approved):

PR #1038, opened 2022-04-04
Do not merge yet: AppenderListener builder proposal

minborg:
API to allow easy creation of useful `AppenderListener` Accumulations using a builder pattern. 

This is just the under-carriage that allows more advanced Reductions to be plugged in like "indexed queues" that can act as a map (storing a key and an associated queue index, upon calling get(key) the method retrieves the index and seeks a tailer to that index after which a value extractor is applied). These more advanced Reductions will reside in CQE (e.g. QBM like structures via index lookups).

Here is an example of how standard Java classes can be used:

```
Reduction<Map<String, MarketData>> latestProtected = Reductions.of(
        ExcerptExtractor.builder(MarketData.class)
                .withMethod(MarketDataProvider.class, MarketDataProvider::marketData)
                .build(),
        Collectors.collectingAndThen(
                Collectors.toConcurrentMap(
                        MarketData::symbol,
                        Function.identity(),
                        replacingMerger()),
                Collections::unmodifiableMap
        )
);
...
MarketData latestAppleMarketData = latest.reduction().get("AAPL");

// Continious and concurrent view
Map<String, MarketData> liveQueueBackedMap = latest.reduction(); 

```
This uses an underlying `ConcurrentHashMap` but exposes an unmodifiable version of the underlying map to protect from inadvertent user modification (i.e. the queue is the single source of truth).

Here is a more rudimentary example where we just record the latest index seen:

```
Reduction<LongSupplier> lastSeeing =
        Reductions.reducingLong(extractingIndex(), 0, (a, b) -> b); 
...
long lastSeen = lastSeeing.reduction().getAsLong();

```

See unit tests for more examples:


nicktindall:
This is very neat

RobAustin (approved):

minborg:
This has been moved to Wire so I am closing this PR now.

PR #1040, opened 2022-04-06
Fix broken test

nicktindall:
This test used to test that you couldn't call `a.setAllNullFields(b)` where `a` and `b` were SingleChronicleQueueBuilders from different hierarchies. It hasn't done that since `wire.generate.tuples` was introduced and set to false by default (the expected exception was being thrown from `Wires.tupleFor`)

As part of another fix I made `Wires.tupleFor` return null and warn when `wire.generate.tuples` is false. That change caused this test to break in a different way.

BTW I checked, and this is the only use of `Wires.tupleFor` outside of `Chronicle-Wire` (in any of our organisations), changing behaviour to return null instead of throwing IllegalArgumentException is breaking I suppose, so would be OK if we want to go back and revert that behaviour change. This PR is still valid though even if we do that.


alamar:
I think the test checked for an interesting use cares, but that of Wire instead of Queue. If we decide to restore that behavior it should be tested in Wire. Approved.

alamar (approved):

PR #1049, opened 2022-04-13
Make Bytes, BytesIn and ByteOut generic to fix raw type warning https://github.com/OpenHFT/Chronicle-Bytes/issues/358

JerryShea:


chronicle-sandbox/src/main/java/net/openhft/chronicle/engine/client/internal/ClientWiredExcerptAppenderStateless.java
```diff
@@ -15,15 +15,15 @@
...
+    private final Bytes<?> source = Bytes.elasticByteBuffer();
     private final Wire wire;
     private ChronicleQueue queue;
     private long cid;
     private long lastWrittenIndex = -1;
 
     public ClientWiredExcerptAppenderStateless(ClientWiredChronicleQueueStateless queue,
                                                ClientWiredStatelessTcpConnectionHub hub,
-                                               Function<Bytes, Wire> wireWrapper) {
+                                               Function<Bytes<?>, Wire> wireWrapper) {
```
minborg, line 26: Not related to this PR but strictly we should declare `Function<? super Bytes<?>, ? extends Wire>`

minborg (approved):
Remove <?> from comments and javadoc in some cases

MigratingToV4.adoc
```diff
@@ -99,9 +99,9 @@ try (DocumentContext dc = appender.writingDocument()) {
 }
 ----
 
-=== Changes to the Bytes API.
+=== Changes to the Bytes<?> API.
```
JerryShea: I think we have to a manual pass through the docs to undo a few of these changes

chronicle-sandbox/src/main/java/net/openhft/chronicle/engine/client/internal/ClientWiredExcerptAppenderStateless.java
peter-lawrey: Possibly, however I am not sure it's clearer

PR #1055, opened 2022-04-22
De-finalize marshallable fields OpenHFT/Chronicle-Wire#414

alamar:
OpenHFT/Chronicle-Wire#414


src/test/java/net/openhft/chronicle/queue/MethodReaderObjectReuseTest.java
```diff
@@ -60,7 +60,7 @@ interface Pinger {
 
     static class PingDTO extends SelfDescribingMarshallable {
         static int constructionCounter, constructionExpected;
-        final Bytes<?> bytes = Bytes.allocateElasticOnHeap();
+        Bytes<?> bytes = Bytes.allocateElasticOnHeap();
```
minborg: Not sure how Bytes works. It is mutable but may have size constraints. If a `final Byte<?>` is declared, one must make sure it can hold any conceivable value in the application at hand?

PR #1064, opened 2022-05-03
improve doco regarding recordHistory

JerryShea:


src/main/java/net/openhft/chronicle/queue/reader/ChronicleHistoryReader.java
```diff
@@ -225,7 +225,7 @@ protected WireParselet parselet() {
     protected void processMessage(CharSequence methodName, MessageHistory history) {
         CharSequence extraHistoId = histosByMethod ? ("_" + methodName) : "";
         long lastTime = 0;
-        // if the tailer has recordHistory(true) then the MessageHistory will be
+        // if the tailer has recordHistory (sourceId != 0) then the MessageHistory will be
```
nicktindall, line 228: should we be removing `recordHistory` altogether here?

nicktindall (approved):

src/main/java/net/openhft/chronicle/queue/reader/ChronicleHistoryReader.java
JerryShea: There is still a getter for recordHistory() (it returns sourceId() != 0)

foroughgoudarzi (approved):

PR #1066, opened 2022-05-03
Limit overlapSize. Fix #1062.

### LINKED ISSUE #1062: Limit the overlapping region to 1 << 30 bytes

foroughgoudarzi:
Limits the overlap size of mapping memory to 1 GB.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -154,6 +154,7 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
             fileAbsolutePath = path.getAbsolutePath();
             wireType = builder.wireType();
             blockSize = builder.blockSize();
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -154,7 +154,7 @@ protected SingleChronicleQueue(@NotNull final SingleChronicleQueueBuilder builde
             fileAbsolutePath = path.getAbsolutePath();
             wireType = builder.wireType();
             blockSize = builder.blockSize();
-            overlapSize = Math.max(64 << 10, builder.blockSize() / 4);
+            overlapSize = Math.min(Math.max(64 << 10, builder.blockSize() / 4), 1L << 30);
```
minborg, line 158: `1L << 30` could be a constant or returned by a method.

JerryShea: +1 for constant

rogersimmons (approved):

PR #1067, opened 2022-05-03
Deprecate useSparseFiles() and sparseCapacity. Fix #1063.

### LINKED ISSUE #1063: Deprecate useSparseFiles() and sparseCapacity() methods

foroughgoudarzi:
The sparse files are used by default if the OS supports it and the sparse capacity is set in the program.


minborg:
Does this mean sparse files will always be used on platforms supporting sparse files?

foroughgoudarzi:
> 

I think so.

rogersimmons (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -874,7 +872,7 @@ public long bufferCapacity() {
     MappedFile mappedFile(File file) throws FileNotFoundException {
         long chunkSize = OS.pageAlign(blockSize);
         long overlapSize = OS.pageAlign(Math.min(blockSize / 4, 1L << 30));
-        return useSparseFile
+        return OS.isSparseFileSupported()
```
rogersimmons: This will always be true on Linux (change in behaviour)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -321,12 +322,11 @@ public WireStoreFactory storeFactory() {
     public SingleChronicleQueue build() {
         boolean needEnterprise = checkEnterpriseFeaturesRequested();
         preBuild();
-        if (Boolean.TRUE.equals(useSparseFiles) && sparseCapacity == null &&
+        if (OS.isSparseFileSupported() &&
```
rogersimmons: This will always be true on Linux

src/test/java/net/openhft/chronicle/queue/ChunkCountTest.java
```diff
@@ -24,7 +25,7 @@ public void chunks() {
                     pos = dc.wire().bytes().writePosition();
                     dc.wire().bytes().writeSkip(16000);
                 }
-                final long expected = builder.useSparseFiles() ? 1 : 1 + (pos >> 18);
+                final long expected = OS.isSparseFileSupported() ? 1 : 1 + (pos >> 18);
```
rogersimmons: Change in behaviour

src/test/java/net/openhft/chronicle/queue/QueueWriteDocumentContextTest.java
```diff
@@ -201,7 +202,7 @@ protected SingleChronicleQueue createQueue(String s) {
                 .timeProvider(new SetTimeProvider("2020/10/19T01:01:01"))
                 .testBlockSize();
         final SingleChronicleQueue queue = builder.build();
-        useSparseFiles = builder.useSparseFiles();
+        useSparseFiles = OS.isSparseFileSupported();
```
rogersimmons: Break

src/test/java/net/openhft/chronicle/queue/WriteBytesTest.java
```diff
@@ -83,7 +84,7 @@ public void testWriteBytesAndDump() {
                 .timeProvider(new SetTimeProvider("2020/10/19T01:01:01"));
         try (ChronicleQueue queue = builder
                 .build()) {
-            final boolean useSparseFiles = builder.useSparseFiles();
+            final boolean useSparseFiles = OS.isSparseFileSupported();
```
rogersimmons: Break

rogersimmons (changes requested):

rogersimmons:
The change as currently made alters the behaviour (dramatically).
The previous useSpareFiles build option defaulted false and was only set true in very rare cases. This change replaced that with OS.isSpareFileSupported which will always be true on Linux - so we end up always using sparse files, rather than only rarely.

The change here is:
1. deprecate the old useSparseFiles variable and any interface which uses them (you have done this)
2. The various bits of code which sample useSparseFiles should be left as they currently are (ie they will activate only if useSparseFiles is true)
3. When we eventually remove useSparseFiles (2 versions after deprecating) any code which follows the useSparseFiles=true path should be removed
4. The related changes to support capped blockSizes is already done

minborg:
> The change as currently made alters the behaviour (dramatically). The previous useSpareFiles build option defaulted false and was only set true in very rare cases. This change replaced that with OS.isSpareFileSupported which will always be true on Linux - so we end up always using sparse files, rather than only rarely.

This is precisely what I said so either we are both wrong and the entire thing is unclear or we are right and there is a fundamental behavioral change introduced. In either case, it is a bad thing.

rogersimmons:
Hi @minborg - there is no fundamental issue here. This change is to deprecate (and eventually remove) the need for users to explicitly call options to instigate the use of sparse files. Instead, the user should use the standard queue interface as usual (normal blockSize etc), then if they use the queue in a sparse manner everything just works transparently. The OS will handle the sparseness under the covers.

This is different to just switching the old sparseFiles flag to default true (from false) as this alters the memory-mapping philosophy to always use one large file rather than blocksizes, which in turn puts excessive virtual memory load on the process.

See the below example which maps a queue normally, but uses it in a sparse manner. Nothing needs to be set for this to work - it just works out of the box transparently. Although the cq4 file appears huge, only 16kB is used.

```
public void sparse() {
        try( ChronicleQueue queue = SingleChronicleQueueBuilder.binary("/dev/shm/large")
                .blockSize(64 << 20)
                .build()) {

            ExcerptAppender appender = queue.acquireAppender();

            try(DocumentContext dc = appender.writingDocument()) {
                dc.wire().bytes().writeInt(64L << 30, 123);
            }
        }
    }
```

```
rw-r--r-- 1 user users 68803362816 May  5 16:51 20220505F.cq4
rw-r--r-- 1 user users      131072 May  5 16:51 metadata.cq4t
```

```
$> hexdump -C 20220505F.cq4 
00000000  c0 00 00 40 b9 06 68 65  61 64 65 72 b6 08 53 43  |...@..header..SC|
00000010  51 53 74 6f 72 65 82 a9  00 00 00 cd 77 72 69 74  |QStore......writ|
00000020  65 50 6f 73 69 74 69 6f  6e 8e 01 00 00 00 00 8d  |ePosition.......|
00000030  02 00 00 00 00 00 00 00  02 00 00 00 00 00 00 00  |................|
00000040  08 01 01 00 00 00 00 00  00 00 00 00 08 01 01 00  |................|
00000050  c8 69 6e 64 65 78 69 6e  67 b6 0c 53 43 51 53 49  |.indexing..SCQSI|
00000060  6e 64 65 78 69 6e 67 82  4c 00 00 00 ca 69 6e 64  |ndexing.L....ind|
00000070  65 78 43 6f 75 6e 74 a5  00 10 cc 69 6e 64 65 78  |exCount....index|
00000080  53 70 61 63 69 6e 67 a5  00 01 cb 69 6e 64 65 78  |Spacing....index|
00000090  32 49 6e 64 65 78 8f a7  c4 00 00 00 00 00 00 00  |2Index..........|
000000a0  c9 6c 61 73 74 49 6e 64  65 78 8e 00 00 00 00 a7  |.lastIndex......|
000000b0  00 01 00 00 00 00 00 00  ca 64 61 74 61 46 6f 72  |.........dataFor|
000000c0  6d 61 74 01 20 80 00 40  b9 0b 69 6e 64 65 78 32  |mat. ..@..index2|
000000d0  69 6e 64 65 78 8f 8f 8d  00 10 00 00 00 00 00 00  |index...........|
000000e0  01 00 00 00 00 00 00 00  e8 80 00 00 00 00 00 00  |................|
000000f0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
000080e0  00 00 00 00 00 00 00 00  1c 80 00 40 b9 05 69 6e  |...........@..in|
000080f0  64 65 78 8f 8f 8f 8f 8d  00 10 00 00 00 00 00 00  |dex.............|
00008100  01 00 00 00 00 00 00 00  08 01 01 00 00 00 00 00  |................|
00008110  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00010100  00 00 00 00 00 00 00 00  01 00 00 00 8f 00 00 00  |................|
00010110  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
1000000000  7b 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |{...............|
1000000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
1005000000
```

```
$> du -h 20220505F.cq4 
16K     20220505F.cq4
```

foroughgoudarzi:
@peter-k-lawrey and @rogersimmons I pushed new changes. This time all the tests passed and the build was successful. @peter-lawrey  I hope I understood what you said today correctly: (Sparse files are used) if and only if (OS allows and blockSize has not been set) based on this I updated the tests too. 
Please take a look.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -874,7 +875,7 @@ public long bufferCapacity() {
     MappedFile mappedFile(File file) throws FileNotFoundException {
         long chunkSize = OS.pageAlign(blockSize);
         long overlapSize = OS.pageAlign(Math.min(blockSize / 4, 1L << 30));
-        return useSparseFile
+        return blockSize == 0
```
rogersimmons: revert (to return useSparseFiles)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -666,13 +667,14 @@ public SingleChronicleQueueBuilder blockSize(int blockSize) {
      */
     public long blockSize() {
 
-        long bs = blockSize == null ? OS.is64Bit() ? 64L << 20 : QueueUtil.testBlockSize() : blockSize;
+       // long bs = blockSize == null ? OS.is64Bit() ? 64L << 20 : QueueUtil.testBlockSize() : blockSize;
```
rogersimmons: revert (uncomment)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -666,13 +667,14 @@ public SingleChronicleQueueBuilder blockSize(int blockSize) {
...
 
         // can add an index2index & an index in one go.
         long minSize = Math.max(QueueUtil.testBlockSize(), 32L * indexCount());
-        return Math.max(minSize, bs);
+        return blockSize != null ? Math.max(minSize, blockSize) : 0;
```
rogersimmons: revert (return Math.max(minSize, bs); )

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -681,19 +683,19 @@ public SingleChronicleQueueBuilder useSparseFiles(boolean useSparseFiles) {
...
         return this;
     }
 
     public long sparseCapacity() {
-        long bs = sparseCapacity == null ? DEFAULT_SPARSE_CAPACITY : sparseCapacity;
-
         // can add an index2index & an index in one go.
         long minSize = Math.max(QueueUtil.testBlockSize(), 64L * indexCount());
-        return Math.max(minSize, bs);
+        return Math.max(minSize, DEFAULT_SPARSE_CAPACITY);
```
rogersimmons: revert (lines 690, 691, and 694. remove 695)

src/test/java/net/openhft/chronicle/queue/ChunkCountTest.java
```diff
@@ -24,7 +25,7 @@ public void chunks() {
                     pos = dc.wire().bytes().writePosition();
                     dc.wire().bytes().writeSkip(16000);
                 }
-                final long expected = builder.useSparseFiles() ? 1 : 1 + (pos >> 18);
+                final long expected = OS.isSparseFileSupported() && builder.blockSize() == 0 ? 1 : 1 + (pos >> 18);
```
rogersimmons: revert (to final... builder.useSparseFiles()...)

src/test/java/net/openhft/chronicle/queue/QueueWriteDocumentContextTest.java
```diff
@@ -201,7 +202,7 @@ protected SingleChronicleQueue createQueue(String s) {
                 .timeProvider(new SetTimeProvider("2020/10/19T01:01:01"))
                 .testBlockSize();
         final SingleChronicleQueue queue = builder.build();
-        useSparseFiles = builder.useSparseFiles();
+        useSparseFiles = OS.isSparseFileSupported() && builder.blockSize() == 0;
```
rogersimmons: revert (to use... = builder.useSparseFiles(); )

src/test/java/net/openhft/chronicle/queue/WriteBytesTest.java
```diff
@@ -80,10 +81,11 @@ public void testWriteBytesAndDump() {
         final SingleChronicleQueueBuilder builder = binary(dir)
                 .testBlockSize()
                 .rollCycle(TEST4_DAILY)
+                .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert (remove)

src/test/java/net/openhft/chronicle/queue/WriteBytesTest.java
```diff
@@ -80,10 +81,11 @@ public void testWriteBytesAndDump() {
...
                 .timeProvider(new SetTimeProvider("2020/10/19T01:01:01"));
         try (ChronicleQueue queue = builder
                 .build()) {
-            final boolean useSparseFiles = builder.useSparseFiles();
+            final boolean useSparseFiles = OS.isSparseFileSupported() && builder.blockSize() == 0;
```
rogersimmons: revert (to builder.useSparseFiles())

src/test/java/net/openhft/chronicle/queue/cleanup/OnReleaseTest.java
```diff
@@ -29,6 +29,7 @@ public void onRelease0() {
                 .binary(path)
                 .rollCycle(RollCycles.MINUTELY)
                 .timeProvider(stp)
+                .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert (remove)

src/test/java/net/openhft/chronicle/queue/cleanup/OnReleaseTest.java
```diff
@@ -38,6 +39,7 @@ public void onRelease0() {
                      .binary(path)
                      .rollCycle(RollCycles.MINUTELY)
                      .timeProvider(stp)
+                     .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert (remove)

src/test/java/net/openhft/chronicle/queue/impl/single/MicroToucherTest.java
```diff
@@ -20,7 +20,7 @@ public class MicroToucherTest extends QueueTestCommon {
     @Test
     public void touchPageSparse() {
         assumeTrue(OS.isLinux());
-        touchPage(b -> b.useSparseFiles(true).rollCycle(RollCycles.HUGE_DAILY), 66561);
+        touchPage(b -> b.rollCycle(RollCycles.HUGE_DAILY), 66561);
```
rogersimmons: revert (to b.useSparseFiles(true)...)

src/test/java/net/openhft/chronicle/queue/impl/single/SparseBinarySearchTest.java
```diff
@@ -64,6 +65,7 @@ public void runWithTimeParameters(RollCycle rollCycle, long incrementInMillis) t
         try (SingleChronicleQueue queue = ChronicleQueue.singleBuilder(getTmpDir())
                 .rollCycle(rollCycle)
                 .timeProvider(stp)
+                .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert

src/test/java/net/openhft/chronicle/queue/impl/single/StoreAppenderInternalWriteBytesTest.java
```diff
@@ -204,6 +205,7 @@ private SingleChronicleQueue createQueue(Path queueDir, TimeProvider timeProvide
                 .binary(queueDir)
                 .rollCycle(RollCycles.TEST4_SECONDLY)
                 .timeProvider(timeProvider)
+                .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert

src/test/java/net/openhft/chronicle/queue/impl/single/TestBinarySearch.java
```diff
@@ -46,6 +47,7 @@ public void testBinarySearch() throws ParseException {
         try (SingleChronicleQueue queue = ChronicleQueue.singleBuilder(getTmpDir())
                 .rollCycle(RollCycles.TEST_SECONDLY)
                 .timeProvider(stp)
+                .blockSize(OS.isWindows() ? 64 << 10 : OS.pageSize())
```
rogersimmons: revert

rogersimmons (changes requested):
Hi @Forough - you're trying to do too much at this stage. See my earlier comment on May 5th.
All you need to do here is:
1. mark as deprecated the old useSparseFiles variables/interface
2. Leave the (now deprecated) sparse file functionality as is. Nothing to change here.
3. Eventually (not now - 2 versions time) we can then remove the code which depends on useSparseFiles
4. Add a cap on blockSizes (already done)

So from here, you just need to revert most of the changes - as highlighted.
The rest of the changes are for a later date.

rogersimmons (approved):

PR #1069, opened 2022-05-04
no longer accepts null for interface name in asMethodReader, Fix #1052

### LINKED ISSUE #1052: Address // TODO: x.23 comments

vach-chronicle:


src/main/java/net/openhft/chronicle/queue/reader/Reader.java
```diff
@@ -55,11 +55,10 @@ public interface Reader {
 
     /**
      * specify method reader interface to use
-     * @param methodReaderInterface interface class name. If null, a dummy reader is created
-     *                              TODO: x.23 this argument will become @NonNull
```
JerryShea, line 59: should document that empty means dummy reader

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/reader/Reader.java
vach-chronicle: done

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -276,8 +276,8 @@ public ChronicleReader historyRecords(final long maxHistoryRecords) {
         return this;
     }
 
-    public ChronicleReader asMethodReader(@Nullable String methodReaderInterface) {
-        if (methodReaderInterface == null)
+    public ChronicleReader asMethodReader(@NotNull String methodReaderInterface) {
+        if (methodReaderInterface.isEmpty())
             entryHandlerFactory = () -> new InternalDummyMethodReaderQueueEntryHandler(wireType);
```
nicktindall, line 281: I wonder if we should throw if someone passes in null or blank? I'm not clear on what the use of `InternalDummyMethodReaderQueueEntryHandler` is?

nicktindall: Or just let it go through to the `Class.forName()` and it'll throw either an NPE or ClassNotFoundException. Seems like garbage-in to me.

nicktindall: I wonder why we take an interface as a string in the first place? surely we could do the Class.forName in the CLI wrapper

vach-chronicle: Its only usage is in tests, it simply reads the binary into a string.
I tried to remove it the first time I approached it, but somehow i was not able to replace it with one simple interface that just converts binary into string. It might be something trivial, but after few attempts i decided to go explore how the wire works.

@JerryShea do you think its better to remove it? if yes i can definitely revisit this, or at least for now I can leave a todo for few versions don't the line to return to it...

JerryShea: See
```
            final String r = commandLine.getOptionValue('r');
            chronicleReader.asMethodReader(r.equals("null") ? null : r);
```
You will need to change this, although I honestly do not know why we have the `InternalDummyMethodReaderQueueEntryHandler` - @peter-lawrey created it originally in 818d9e90faa194c7fa27f264f961a64898115fa7. To me it seems logical that a method reader has to be for an interface

vach-chronicle: ok addressed the null,
i'll check why we need the dummy reader over the weekend, I tried to remove it at first...

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
vach-chronicle: Ok i'll merge this for now and get rid of the nullable
but I'll address the InternalDummyMethodReaderQueueEntryHandler later @nicktindall 
just need to ask Peter

PR #1070, opened 2022-05-05
Configure the binary compatibility checker, Fix #1060

### LINKED ISSUE #1060: The binary compatibility tool does not check the package net.openhft.chronicle.queue.impl
Packages matching "*impl*" and "*internal*" are not checked. Yet, there are public methods in the "*impl*" package.

The root problem is that public classes reside in an "impl" package (like `SingleChronicleQueueBuilder` ) and these should be migrated out and they should be split up into interfaces (in the public API) and implementing classes (in the internal package).

minborg:
This PR will make the compatibility checker check `impl` as this, unfortunately, is a part of the public API.

We should fix this someday.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -311,6 +311,7 @@
             <plugin>
                 <groupId>net.openhft</groupId>
                 <artifactId>binary-compatibility-enforcer-plugin</artifactId>
+                <version>1.0.13</version>
                 <executions>
                     <execution>
                         <phase>verify</phase>
@@ -321,7 +322,17 @@
                             <referenceVersion>5.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
                             <!-- Reduce the API surface of ChronicleReader -->
-                            <binaryCompatibilityPercentageRequired>86.1</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>89.3</binaryCompatibilityPercentageRequired>
+                            <extraOptions>
+                                <!--This skips "internal" but checks "impl"-->
+                                <extraOption>
+                                    <name>-keep-internal</name>
+                                </extraOption>
+                                <extraOption>
+                                    <name>skip-internal-packages</name>
+                                    <value>.*internal.*</value>
+                                </extraOption>
+                            </extraOptions>
                         </configuration>
                     </execution>
                 </executions>


pom.xml
```diff
@@ -321,7 +322,17 @@
...
-                            <binaryCompatibilityPercentageRequired>86.1</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>89.3</binaryCompatibilityPercentageRequired>
+                            <extraOptions>
+                                <!--This skips "internal" but checks "impl"-->
+                                <extraOption>
+                                    <name>-keep-internal</name>
+                                </extraOption>
+                                <extraOption>
+                                    <name>skip-internal-packages</name>
+                                    <value>.*internal.*</value>
```
nicktindall, line 333: Should this have a hyphen in front of it too? `-keep-internal` and `-skip-internal-packages` both seem to have hyphens in `japi-compliance-checker -help`

```
  -keep-internal
      Do not skip checking of these packages:
        *impl*
        *internal*
        *examples*
  
  -skip-internal-packages PATTERN
      Do not check packages matched by the regular expression.
  
  -skip-internal-types PATTERN
      Do not check types (classes and interfaces) matched by the regular
      expression. It's matched against full qualified type names (e.g.
      'org.xyz.Name<T>'). It has to match any part of type name.
```

nicktindall: Also should this config be in pluginManagement in java-parent-pom perhaps?

nicktindall (approved):
Looks good, will leave it up to you as to whether to make changes in response to my comments

alamar:
I suggest getting away from trying to interpret parameters of `japi-compliance-checker` and just pass it as a list:

```
                                <!--This skips "internal" but checks "impl"-->
                                <params>
                                    <param>-keep-internal</param>
                                    <param>-skip-internal-packages</param>
                                    <param>.*internal.*</param>
                                </params>
```
Check what's native MOJO XML API for Lists.

pom.xml
minborg: No `-` in the beginning.

minborg:
> I suggest getting away from trying to interpret parameters of `japi-compliance-checker` and just pass it as a list:
> 
> ```
>                                 
>                                 <params>
>                                     <param>-keep-internal</param>
>                                     <param>-skip-internal-packages</param>
>                                     <param>.*internal.*</param>
>                                 </params>
> ```
> 
> Check what's native MOJO XML API for Lists.

https://github.com/OpenHFT/Binary-Compatibility-Enforcer-Plugin/issues/10

PR #1072, opened 2022-05-08
Queue performance using sparse file and chunking. Fix #1068.

### LINKED ISSUE #1068: Add a benchmark to compare performance using chunking and a single region.
@foroughgoudarzi please add a description. 
The benchmark compares queue performances when memory mapping is using a large sparse file,(1<<40) a large block size (1<<40) and a standard block size (64L << 20).

foroughgoudarzi:


src/test/java/net/openhft/chronicle/queue/bench/QueueSparseFilesJLBHBenchmark.java
```diff
@@ -0,0 +1,109 @@
...
+        tailer = sinkQueue.createTailer().disableThreadSafetyCheck(true);
+        this.jlbh = jlbh;
+        writeProbe = round == 1 ? jlbh.addProbe("write (sparse file)") : jlbh.addProbe("write (chunking)");
+        readProbe = round == 1 ? jlbh.addProbe("read (sparse file)") : jlbh.addProbe("read (chunking)");
+    }
+
+    @Override
+    public void run(long startTimeNS) {
+
+        try (DocumentContext dc = appender.writingDocument()) {
```
rogersimmons, line 93: for sparse files, probably a more suitable test would be to write a message every so often - eg every 1MB, or every 64MB etc, rather than consecutive messages.
To do this, use eg 
````
dc.wire().bytes().writeInt(pos, value); // eg ...writeInt(64L<<20, 1234) to write 1234 at the 64MB point
````
So eg add a long index which starts at 0 then increments by 1L<<20 every write.

Will need to choose a reasonable step size so can have enough points to sample without going too huge on the file. A sparse file of ~1TB containing a message every 1MB for 1M messages would be ok for example (or every 100k for 10M messages)

Also, add a round-3 test which uses a standard blocksize (eg 256L<<20) to compare with the giant block size

rogersimmons (changes requested):

src/test/java/net/openhft/chronicle/queue/bench/QueueSparseFilesJLBHBenchmark.java
foroughgoudarzi: Now the message size is 1k bytes but the content is only 11 bytes so I think it resembles sparse files.

foroughgoudarzi:
Closed as it mixed with another branch. Created new PR in https://github.com/OpenHFT/Chronicle-Queue/pull/1074

PR #1081, opened 2022-05-11
File shrinkage

RobAustin:
the test fails.. lets chat about this.


alamar:
I think we got rid of file shrinking and are not ready to reintroduce it right now.

It was implemented as an afterthought and was dragged as non-functional, non-tested feature until I removed it a month ago.

peter-lawrey:
Closing as obsolete. This PR changes the `QueueFileShrinkManager` subsystem, which was deliberately removed in `19aeb1fc4` after remaining non-functional and untested. Queue no longer owns this cleanup policy; any future roll-maintenance implementation belongs behind the Enterprise integration points, not by restoring the removed implementation to OSS Queue.

Issue #1059, opened 2022-04-27
NPE when closing tailer in parallel


alamar:
One of our test gives
```
java.lang.AssertionError: [ExceptionKey{level=WARN, clazz=class software.chronicle.fix.staticcode.ConnectorEventHandler, message='78b236a0 threw ', throwable=java.lang.NullPointerException
 at net.openhft.chronicle.queue.impl.single.SCQIndexing.linearScan0(SCQIndexing.java:380)
 at net.openhft.chronicle.queue.impl.single.SCQIndexing.linearScan(SCQIndexing.java:338)
 at net.openhft.chronicle.queue.impl.single.SCQIndexing.moveToIndex0(SCQIndexing.java:315)
 at net.openhft.chronicle.queue.impl.single.SCQIndexing.moveToIndex(SCQIndexing.java:254)
 at net.openhft.chronicle.queue.impl.single.SingleChronicleQueueStore.moveToIndexForRead(SingleChronicleQueueStore.java:228)
 at net.openhft.chronicle.queue.impl.single.StoreTailer.moveToIndexResult0(StoreTailer.java:668)
 at net.openhft.chronicle.queue.impl.single.StoreTailer.moveToIndexResult(StoreTailer.java:694)
 at net.openhft.chronicle.queue.impl.single.StoreTailer.originalToEnd(StoreTailer.java:960)
 at net.openhft.chronicle.queue.impl.single.StoreTailer.callOriginalToEnd(StoreTailer.java:874)
 at net.openhft.chronicle.queue.impl.single.StoreTailer.toEnd(StoreTailer.java:856)
 ```

alamar:
See also #1058

alamar:
It is also possible to get NPE in `SCQIndexing.moveToIndex0`, and even a JVM crash (file attached)

[hs_err_pid798701.log](https://github.com/OpenHFT/Chronicle-Queue/files/8572976/hs_err_pid798701.log)

RobAustin:
which version of queue, which line is producing the NPE, the latest develop branch has no code on that line
![image](https://user-images.githubusercontent.com/6506056/167128810-fbbbfa37-c72b-43a8-bdd6-0ba445835f1e.png)

alamar:
Just run `develop` test `TestUninstallableEventLoop#startAndConfirmAllHandlersGone` with JVM arg `-Xint` and around 10th iteration you would see:

* NPE in `linearScan0`
[...]

PR #1090, opened 2022-05-17
Avoid NPE when tailer is closed in parallel

alamar:
fixes OpenHFT/Chronicle-Queue#1059

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
@@ -261,6 +261,9 @@ class SCQIndexing extends AbstractCloseable implements Demarshallable, WriteMars
     private ScanResult moveToIndexFromTheStart(@NotNull ExcerptContext ec, long index) {
         try {
             Wire wire = ec.wire();
+            if (wire == null)
+                return ScanResult.END_OF_FILE;
+
             wire.bytes().readPositionUnlimited(0);
             if (wire.readDataHeader())
                 return linearScan(wire, index, 0, wire.bytes().readPosition());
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -261,6 +261,9 @@ ScanResult moveToIndex(@NotNull final ExcerptContext ec, final long index) {
     private ScanResult moveToIndexFromTheStart(@NotNull ExcerptContext ec, long index) {
         try {
             Wire wire = ec.wire();
+            if (wire == null)
```
JerryShea, line 264: is it possible to add a test that breaks before this change?

alamar: It is `TestUninstallableEventLoop` (on repeat, before the ChronicleEnterprise/Chronicle-FIX#1053)

alamar:
@JerryShea Please check, I've added test.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -645,11 +646,15 @@ private ScanResult moveToIndexResult0(final long index) {
...
         index(index);
         final ScanResult scanResult = this.store().moveToIndexForRead(this, sequenceNumber);
-        final Bytes<?> bytes = privateWire().bytes();
         switch (scanResult) {
             case FOUND:
-                state = FOUND_IN_CYCLE;
-                moveToState.onSuccessfulLookup(index, direction, bytes.readPosition());
+                Wire privateWire = privateWire();
+                if (privateWire == null)
+                    state = END_OF_CYCLE;
```
JerryShea, line 656: we should javadoc TailerState!

JerryShea (approved):
LGTM

peter-lawrey (approved):

PR #1098, opened 2022-05-30
Implement a test for the surefire interrupt flag bug, remove surefire:3.0.0-M4 pinning

nicktindall:
Not sure if we need to merge this, but just a data point for the surefire thing

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/SurefireInterruptFlagTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/SurefireInterruptFlagTest.java
@@ -1,0 +1,21 @@
+package net.openhft.chronicle.queue;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


src/test/java/net/openhft/chronicle/queue/SurefireInterruptFlagTest.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Test for https://issues.apache.org/jira/browse/SUREFIRE-1863
+     */
+    @Test
+    void testSurefireLeavesInterruptFlagIntactOnOutput() {
+        assertFalse(Thread.currentThread().isInterrupted());
+        Thread.currentThread().interrupt();
+        assertTrue(Thread.currentThread().isInterrupted());
+        System.out.println("Hello world!");
+        assertTrue(Thread.currentThread().isInterrupted());
```
nicktindall, line 19: This fails under surefire 3.0.0-M5, but passes under M4 and M6

JerryShea (approved):

PR #1103, opened 2022-06-02
Cache queue and tailer so ChronicleHistoryReader#readChronicle can be called repeatedly. Fixes #1102

### LINKED ISSUE #1102: Cache queue and tailer so ChronicleHistoryReader#readChronicle can be called repeatedly

JerryShea:
#comment `ChronicleHistoryReader` now implements `Closeable`


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -194,7 +194,7 @@ protected SingleChronicleQueue(@NotNull final SingleChronicleQueueBuilder builde
 
             // release the write lock if the process is dead
             if (writeLock instanceof TableStoreWriteLock) {
-                ((TableStoreWriteLock) writeLock).forceUnlockIfProcessIsDead();
```
JerryShea, line 197: clean up

nicktindall (approved):

PR #1111, opened 2022-06-16
Fix ChronicleReader timezone configurability tests

nicktindall:
This test used to work because a new `MilliTimestampLongConverter` used to be created for each ChronicleReader run (and the timezone system property was re-evaluated in the constructor.
Something's changed in the lifecycle of the converter such that it is a singleton created on first use (with a fixed timezone per VM instance). I've just switched to spawning a process with the system property and checking the output is correct rather than doing it in-process.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/internal/reader/ChronicleReaderTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/internal/reader/ChronicleReaderTest.java
@@ -3,7 +3,9 @@ package net.openhft.chronicle.queue.internal.reader;
 import net.openhft.chronicle.bytes.MethodReader;
 import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.OS;
+import net.openhft.chronicle.core.annotation.RequiredForClient;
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/queue/internal/reader/ChronicleReaderTest.java
```diff
@@ -493,23 +489,32 @@ public void configure(Reader reader) {
...
-            }
+        for (Long timestamp : timestamps) {
+            final String expectedTimestamp = mtlc.asString(timestamp);
+            int timestampIndex = output.indexOf(expectedTimestamp);
+            assertTrue(String.format("%s contains %s", output, expectedTimestamp), timestampIndex > 0);
+            output = output.substring(timestampIndex + expectedTimestamp.length());
+        }
+    }
+
+    private static class ChronicleReadeRunner {
```
JerryShea: Reade**r** ?

JerryShea (approved):

nicktindall:
Depends on https://github.com/OpenHFT/Chronicle-Core/pull/414

PR #1113, opened 2022-06-17
Fix LastAppendedTest by ensuring deregistration of history reader

alamar:


src/test/java/net/openhft/chronicle/queue/LastAppendedTest.java
```diff
@@ -13,20 +11,10 @@
...
-   // public void testLastWritten() {
-    //    builder(this::testLastWritten0).build().run();
-   // }
-
-
-   // public void testLastWrittenMetadata() {
-  //      builder(this::testLastWrittenMetadata0).build().run();
-  //  }
     @Test
     void testLastWritten0() {
```
RobAustin: I guess we should change the method names back to "testLastWritten" from "testLastWritten0"

RobAustin (approved):

PR #1120, opened 2022-07-04
reuse last store when applicable

vach-chronicle:


RobAustin (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -948,7 +948,7 @@ void cleanupStoreFilesWithNoData() {
             writeLock.unlock();
             if(fireOnReleasedEvent != null)
                 BackgroundResourceReleaser.run(fireOnReleasedEvent);
-            long tookMillis = (System.nanoTime() - start) / 1_000_000;
+            long tookMillis = (time.currentTimeNanos() - start) / 1_000_000;
```
RobAustin: when reporting the "time took", I don't feel we should use the time provider, In other words, the old code looks more correct to me.

RobAustin (changes requested):
time taken should use System.nano time rather than the time provider

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
JerryShea: Correct. There are a number of places in the code that need real time, not time provider time

RobAustin (approved):

PR #1128, opened 2022-07-20
Improve description of defaultRollCycle

alamar:


src/main/java/net/openhft/chronicle/queue/QueueSystemProperties.java
```diff
@@ -23,8 +23,9 @@ private QueueSystemProperties() {
      * Name of a system property used to specify the default roll cycle.
      * <p>
      * System Property key: "net.openhft.queue.builder.defaultRollCycle"
```
JerryShea, line 25: I know it's not part of this commit but I am wary of duplicate information e.g. `System Property key:`, although I love them idea of this class containing all CQ system properties

JerryShea (approved):

Issue #1073, opened 2022-05-09
Backward tailer not releasing the file when closed


ajax-lizogubenko-s:
Chronicle queue version: 5.22.18

We are experiencing huge problems with files rollover.
We creating one utility tailer with SingleChronicleQueue.createTailer() perform some readText() operation with BACKWARD direction and then closing it with close() method. 
Unfortunately we can see that last file that was read by Tailer stays blocked:
<img width="1302" alt="Screenshot 2022-05-09 at 12 07 25" src="https://user-images.githubusercontent.com/91266605/167378319-fe20f1a2-8900-4207-98c0-196d905eb4bf.png">
So we can't remove old files, cause last one is being locked. And we need to remove old files

File being released only after application restart

UPD: looks like countExcerpts() has the same problem

JerryShea:
@ajax-lizogubenko-s now that #1148 is fixed, can you re-try this with the latest version?

nicktindall:
The problem with `countExcerpts()` will persist in x.23 because it uses `acquireTailer()`. It's fixed in x.24ea but we should backport the changes to stop using `acquireTailer()` to x.23 if we want it to not leave file handles open.

JerryShea:
No response received from original poster, so closing as assuming fixed by #1148

PR #1145, opened 2022-08-24
Add test for file handle leak, fix existing AppenderFileHandleLeakTest

nicktindall:
This is a leak identified by a client. Possibly related to #1073 

The problem seems to be in `ReferenceCountedCache`, it only purges the Xth `MappedFile` when the X+2nd `MappedFile` is retrieved.

queue -> WireStorePool -> WireStoreSupplier -> ReferenceCountedCache (mappedFileCache)

The Xth MappedFile is eligible to be cleaned up by `ReferenceCountedCache#bgCleanup` much earlier but there's nothing to trigger it. It only triggers again on the subsequent `ReferenceCountedCache#get`


JerryShea (approved):
Fantastic

PR #1153, opened 2022-08-30
Attempt to de-flake AppenderFileHandleLeakTest

nicktindall:
Not sure if this will do it, let's let it run a few times to see before we merge

It just waits up to 5 seconds for the leaked file handles to go away. Assuming the flakiness can be explained that way, but not sure. 5 seconds is an eternity so if it still fails it's something else.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
@@ -29,6 +29,8 @@ import net.openhft.chronicle.queue.*;
 import net.openhft.chronicle.queue.impl.StoreFileListener;
 import net.openhft.chronicle.testframework.FlakyTestRunner;
 import net.openhft.chronicle.threads.NamedThreadFactory;
+import net.openhft.chronicle.threads.Pauser;
+import net.openhft.chronicle.threads.TimingPauser;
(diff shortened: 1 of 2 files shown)


JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
```diff
@@ -276,11 +278,29 @@ public void tailerShouldOnlyKeepCurrentRollCycleOpen() {
 
     private void assertOnlyCurrentRollCycleIsOpen(long timestamp) {
         BackgroundResourceReleaser.releasePendingResources();
+        /*
+         * "A mapped byte buffer and the file mapping that it represents remain valid until the buffer itself is garbage-collected."
+         *
+         * Given we can't guarantee a GC happens, I wonder if this test can ever not be flaky
```
JerryShea, line 298: good comment ;) Were you gonna add FlakyTestRunner?

JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
nicktindall: Pushed, let's see if it helps

PR #1168, opened 2022-09-13
Allow earlyAcquireNextCycle by avoiding peek or EOF the future

### LINKED ISSUE #1157: Pretoucher: enabling earlyAcquireNextCycle causes Tailers to skip over messages
When a pre-toucher is running on a queue and `earlyAcquireNextCycle` is enabled, any Tailer reading from it can skip over messages when it hits the end of a roll cycle and sees the next one available.

This is trivially reproducible in https://github.com/OpenHFT/Chronicle-Queue/pull/1156/files and https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/347

alamar:
fixes #1157

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -396,7 +396,7 @@ class StoreAppender extends AbstractCloseable
             writeLock.lock();
             int cycle = queue.cycle();
             if (wire == null)
-                setWireIfNull(cycle);
+                setCycle2(cycle, true);
 
             if (this.cycle != cycle)
                 rollCycleTo(cycle);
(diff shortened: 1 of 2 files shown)


JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -717,7 +695,7 @@ void beforeAppend(final Wire wire, final long index) {
     // throws UnrecoverableTimeoutException
 
     private void rollCycleTo(final int cycle) {
-        rollCycleTo(cycle, false);
+        rollCycleTo(cycle, this.cycle > cycle);
```
JerrySheaB4: comment from @peter-k-lawrey - pleasd consider renaming parameter

PR #1171, opened 2022-09-15
Use /proc/self/maps instead of lsof for AppenderFileHandleLeakTest

nicktindall:
`lsof` seems to block inexplicably sometimes, `pmap` does as well and the output is different on every version/distro. `/proc/self/maps` seems like a much more reliable way to determine mapped files.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/MappedFileUtil.java
+++ b/src/test/java/net/openhft/chronicle/queue/MappedFileUtil.java
@@ -1,0 +1,66 @@
+package net.openhft.chronicle.queue;
+
+import net.openhft.chronicle.core.Jvm;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public enum MappedFileUtil {
+    ;
+    private static final Path PROC_SELF_MAPS = Paths.get("/proc/self/maps");
+
+    // See https://man7.org/linux/man-pages/man5/proc.5.html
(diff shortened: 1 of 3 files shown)


src/test/java/net/openhft/chronicle/queue/MappedFileUtil.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public enum MappedFileUtil {
```
nicktindall, line 16: We could promote this to Chronicle-Test-Framework if it proves reliable. It seems like it might be useful in other contexts.

src/test/java/net/openhft/chronicle/queue/MappedFileUtil.java
```diff
@@ -0,0 +1,66 @@
...
+    ;
+    private static final Path PROC_SELF_MAPS = Paths.get("/proc/self/maps");
+
+    // See https://man7.org/linux/man-pages/man5/proc.5.html
+    private static final Pattern LINE_PATTERN = Pattern.compile("([A-Fa-f\\d\\-]+)\\s+([rwxsp\\-]+)\\s+([A-Fa-f\\d]+)\\s+([A-Fa-f\\d]{2}:[A-Fa-f\\d]{2})\\s+(\\d+)\\s*(.*)?");
```
JerryShea: phew!

JerryShea (approved):

src/test/java/net/openhft/chronicle/queue/MappedFileUtil.java
nicktindall: I love a good regex. A lot of the bulk is the hexadecimal number pattern repeated `[A-Fa-f\d]+`. I swapped it out for the `\p{XDigit}` macro, hopefully that makes it less of a sledge hammer in the face.

PR #1174, opened 2022-09-16
Deprecate Pretoucher

JerryShea:
Deprecate Pretoucher #1177

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
@@ -25,7 +25,6 @@ import net.openhft.chronicle.core.io.AbstractCloseable;
 import net.openhft.chronicle.core.io.ClosedIllegalStateException;
 import net.openhft.chronicle.core.threads.InvalidEventHandlerException;
 import net.openhft.chronicle.core.time.TimeProvider;
-import net.openhft.chronicle.wire.Wire;
 
 import java.util.function.IntConsumer;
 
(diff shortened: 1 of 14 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
```diff
@@ -126,27 +126,12 @@ private void assignCurrentCycle() {
...
-                    }
-                SingleChronicleQueueStore oldStore = currentCycleWireStore;
-                currentCycleWireStore = queue.storeForCycle(qCycle, queue.epoch(), earlyAcquireNextCycle || canWrite, currentCycleWireStore);
-                if (oldStore != null && oldStore != currentCycleWireStore)
-                    oldStore.close();
-            } finally {
-                if (canWrite)
-                    queue.writeLock().unlock();
-            }
+            SingleChronicleQueueStore oldStore = currentCycleWireStore;
```
alamar: After `releaseResources()`, `currentCycleWireStore` will be null, won't it?

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
```diff
@@ -155,12 +140,10 @@ private void assignCurrentCycle() {
                 cycleChangedListener.accept(qCycle);
 
                 if (earlyAcquireNextCycle)
```
alamar, line 141: Why we are only writing a message if this flag is set? I guess this line has to go and that isDebugEnabled the one that got to stay.

alamar:
> RollCycleMultiThreadStressPretouchEATest
> RollCycleMultiThreadStressPretouchTest

Still fail every time when checkEA() is removed - appenders/tailers are apparently not ready for EA

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
JerryShea: good point

JerryShea:
> > RollCycleMultiThreadStressPretouchEATest
> > RollCycleMultiThreadStressPretouchTest
> 
> Still fail every time when checkEA() is removed - appenders/tailers are apparently not ready for EA

Yes - absolutely. To be expected

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
JerryShea: fixed

peter-lawrey (approved):

PR #1175, opened 2022-09-16
Move warning in deprecation message to next release

alamar:


JerryShea (approved):
Agreed. We should have removed at the start of x.23

PR #1180, opened 2022-09-20
Make LOCKING_PROCESS_DEAD the default force unlock mode

### LINKED ISSUE #1179: Make LOCKING_PROCESS_DEAD the default force unlock mode
See #995

alamar:
closes #1179


JerryShea (approved):
LGTM but would like @nicktindall to confirm is OK

nicktindall:
I like it a lot except it may be a headache for our friends using docker without `pid=host`. I guess it should be a very rare occurrence. We should make sure the `pid=host` recommendation is prominent, perhaps even mentioned or linked to in the exception message?

nicktindall (approved):

alamar:
@nicktindall Can you please file a separate ticket/PR about that? I didn't change that part of code so I don't want to grow the footprint here.

PR #1194, opened 2022-09-28
Move Pretoucher to Queue-Enterprise

JerryShea:


README.adoc
```diff
@@ -1554,10 +1552,8 @@ The constructor of Pretoucher takes the name of the queue that this Pretoucher i
 // Creates the queue q1 (or q1 is a queue that already exists)
 try(final SingleChronicleQueue q1 = SingleChronicleQueueBuilder.binary("queue-storage-path").build();
 
-    // Creates Pretoucher pretouch for q1
-    final Pretoucher pretouch = new Pretoucher(q1)){
+    final Pretoucher pretouch = PretouchUtil.createPretoucher(q1)){
```
alamar: Why still have this?

README.adoc
```diff
@@ -1566,13 +1562,14 @@ try(final SingleChronicleQueue q1 = SingleChronicleQueueBuilder.binary("queue-st
 }
 ----
 
-The method shutdown(), closes the queue and releases any other resources.
+The method `close()`, closes the Pretoucher and releases its resources.
```
alamar, line 1565: Why talk about it in CQ? Maybe just mention we have Pretoucher in Enterprise.

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
```diff
@@ -36,130 +32,34 @@
...
-    private final boolean earlyAcquireNextCycle;
-    private final boolean canWrite;
-    private final PretoucherState pretoucherState;
-    private final TimeProvider pretouchTimeProvider;
-    private int currentCycle = Integer.MIN_VALUE;
-    private SingleChronicleQueueStore currentCycleWireStore;
-    private MappedBytes currentCycleMappedBytes;
-
+// TODO: x.25 turn into interface
+public class Pretoucher extends AbstractCloseable {
```
alamar: Perhaps we can get rid of this class at all and use a Closeable Callable?

README.adoc
JerryShea: For reasons of binary compatibility

README.adoc
JerryShea: Yes fair point. If it's OK with you I will see if the doco team can take that one (and update antora docs too)

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
JerryShea: Unfortunately we have to support the same class signature for one more release cycle

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
```diff
@@ -36,130 +32,34 @@
...
-    private final boolean canWrite;
-    private final PretoucherState pretoucherState;
-    private final TimeProvider pretouchTimeProvider;
-    private int currentCycle = Integer.MIN_VALUE;
-    private SingleChronicleQueueStore currentCycleWireStore;
-    private MappedBytes currentCycleMappedBytes;
-
+    @Deprecated(/* remove in x.25 */)
     public Pretoucher(final SingleChronicleQueue queue) {
-        this(queue,
```
nicktindall, line 58: In these constructors we should probably warn that this is a dummy implementation and that it won't do anything. I didn't even know PretouchUtil existed previously so I wonder how many clients create Pretouchers directly using this class.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -269,7 +269,7 @@ static SingleChronicleQueueStore createStore(@NotNull RollingChronicleQueue queu
         return wireStore;
     }
 
-    static boolean isQueueReplicationAvailable() {
+    public static boolean isQueueReplicationAvailable() {
```
nicktindall: Should we make an alias for this `areEnterpriseFeaturesAvailable` and deprecate this one, if we're expanding its use beyond replication (just looks a bit WTF when you're asking that in the context of the pretoucher)

nicktindall: In fact given it wasn't public previously we could probably get away with just renaming it.

src/test/java/net/openhft/chronicle/queue/impl/single/stress/RollCycleMultiThreadStressTest.java
```diff
@@ -165,6 +165,7 @@ public void stress() throws Exception {
         if (PRETOUCH) {
             pretoucherThread = new PretoucherThread(file);
             executorServicePretouch.submit(pretoucherThread);
+            expectException("touchPage failed");
```
nicktindall: Do we still need this?

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
JerryShea: yep - agree

src/test/java/net/openhft/chronicle/queue/impl/single/stress/RollCycleMultiThreadStressTest.java
JerryShea: Yes. RollCycleMultiThreadStressPretouchEATest is now in QE and descends from this

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
JerryShea: Have added a warning in static block

pom.xml
```diff
@@ -604,18 +605,6 @@
                                     </mainClass>
                                 </configuration>
                             </execution>
-                            <execution>
-                                <id>QueueMultiThreadedJLBHBenchmark</id>
```
JerryShea, line 608: moved to QE

nicktindall (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
alamar: I don't think we need to keep it if the feature is removed. Yes, we've marked it late, but already in 2,23

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
```diff
@@ -36,130 +32,38 @@
...
-        } catch (ClassCastException cce) {
-            Jvm.warn().on(getClass(), cce);
-
-        } catch (IllegalStateException e) {
-            if (queue.isClosed())
-                throw new InvalidEventHandlerException(e);
-            else
-                Jvm.warn().on(getClass(), e);
-        }
+        // do nothing
```
alamar: I think it should at least produce a warning if used.

src/main/java/net/openhft/chronicle/queue/util/PretouchUtil.java
```diff
@@ -18,16 +18,49 @@
...
+    private static final Function<SingleChronicleQueue, Pretoucher> PRETOUCHER_FACTORY;
+    private static final Function<SingleChronicleQueue, EventHandler> EH_FACTORY;
 
-    private PretouchUtil() {}
+    static {
+        Function<SingleChronicleQueue, Pretoucher> ptFactory;
+        Function<SingleChronicleQueue, EventHandler> ehFactory;
+        try {
+            final Class<?> clazz = Class.forName("software.chronicle.enterprise.queue.pretoucher.EnterprisePretouchUtil");
+            final Method createPretoucher = Jvm.getMethod(clazz, "createPretoucher", SingleChronicleQueue.class);
```
alamar: Maybe we could convert Pretoucher to an interface and use it? Only cast it once instead of calling numerous reflected methods. That's fragile.

alamar (changes requested):

src/main/java/net/openhft/chronicle/queue/impl/single/Pretoucher.java
JerryShea: It will - see static block in Pretoucher

src/main/java/net/openhft/chronicle/queue/util/PretouchUtil.java
JerryShea: Yes it is horrible, but I couldn't make Pretoucher an interface without breaking binary compatibility.

But - you know what - we said it was gonna be deprecated in x.24 so we *can* make it an interface - I'll do that, and break binary compatibility (that was my first solution but I got cold feet)

src/main/java/net/openhft/chronicle/queue/util/PretouchUtil.java
```diff
@@ -19,15 +19,46 @@
...
+public final class PretouchUtil implements PretoucherFactory {
+    public static final PretoucherFactory INSTANCE;
 
-    private PretouchUtil() {}
+    static {
+        PretoucherFactory instance;
+        try {
+            final Class<?> clazz = Class.forName("software.chronicle.enterprise.queue.pretoucher.EnterprisePretouchUtil");
+            instance = (PretoucherFactory) ObjectUtils.newInstance(clazz);
+            assert SingleChronicleQueueBuilder.areEnterpriseFeaturesAvailable();
```
alamar, line 37: Will not happen unless assertions are enabled - does we depend on this check or is it nice to have?

alamar (approved):

src/main/java/net/openhft/chronicle/queue/util/PretouchUtil.java
JerryShea: just a nice to have - really for our internal benefit

PR #1195, opened 2022-09-28
Feature/stabilise pretoucher 23

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/PretoucherState.java
```diff
@@ -104,16 +104,23 @@ public void pretouch(MappedBytes bytes) {
                         // ignored.
                     }
                     try {
+                        if (bytes != null && lastTouchedPage >= realCapacity) {
```
alamar, line 107: I suggest not doing any of these changes in Chronicle-Queue repository, rather moving these classes as is to CQE and then changing the classes there. What do you think?

JerryShea: This has now been replaced by another PR

PR #1199, opened 2022-09-28
Feature/combine test base classes

JerryShea:
@peter-lawrey further to our conversation about tests behaving more consistently in IntelliJ vs maven - this definitely helps


src/test/java/net/openhft/chronicle/queue/impl/single/RollingCycleTest.java
```diff
@@ -57,11 +56,6 @@ public static Collection<Object[]> data() {
         );
     }
 
-    @BeforeClass
-    public static void sync() {
-        System.setProperty("chronicle.queue.synchronousFileShrinking", "true");
```
JerryShea, line 62: duplicated code

nicktindall (approved):

Issue #1197, opened 2022-09-28
Found non-matching line in /proc/self/maps


alamar:
When running tests on local Linux machine, getting a lot of messages like these:

```
[main] WARN net.openhft.chronicle.queue.MappedFileUtil - Found non-matching line in /proc/self/maps: 7f3b9ee38000-7f3b9ee3b000 r--p 00000000 103:03 3147723                   /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
```

hft-team-city:
Released in [Chronicle-Queue-5.23.36](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.23.36), [BOM-2.23.127](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23.127)

PR #1200, opened 2022-09-29
Fix and test the regex used to interpret /proc/self/maps lines

nicktindall:


JerryShea:
Fixes https://github.com/OpenHFT/Chronicle-Queue/issues/1197 ?

JerryShea (approved):

PR #1207, opened 2022-10-04
Don't track debug messages in ExceptionTracker

nicktindall:
That test that was spewing out debug messages in CQE was actually extending this base class.


src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
```diff
@@ -94,8 +93,6 @@ public void dontOverwriteExisting() {
 
     @Test
     public void dontOverwriteExistingDifferentQueueInstance() {
-        expectException("Trying to overwrite index 0 which is before the end of the queue");
-        expectException("Trying to overwrite index 1 which is before the end of the queue");
```
nicktindall, line 98: This got downgraded to a debug message at some point

JerryShea (approved):

PR #1213, opened 2022-10-09
restore previous signature for binary compatibility

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/BinarySearch.java
```diff
@@ -32,6 +32,13 @@
 public enum BinarySearch {
     INSTANCE;
 
+    @Deprecated(/* Remove in x.25. Use search(net.openhft.chronicle.queue.ExcerptTailer, net.openhft.chronicle.wire.Wire, java.util.Comparator<net.openhft.chronicle.wire.Wire>) */)
```
alamar: What makes us think x.25 will actually see it removed? Maybe clarify usage in the comment.

nicktindall (approved):

alamar (approved):

PR #1223, opened 2022-10-14
Deprecate esoteric roll cycles, move them to separate enums

### LINKED ISSUE #1221: Put esoteric/test/legacy roll cycles a little more out-of-reach
The RollCycles enum contains definitions for a bunch of common roll cycles. It also includes some definitions for legacy roll cycles, test roll cycles and some specialised roll cycles for unusual configurations.

Currently if you're looking at the decompiled RollCycles class there's no indication of which ones you should probably use (it's explained in JavaDoc) and getting it wrong can have some unexpected results.

The plan is to leave the "common"/"sensible" set of roll cycles in the RollCycles enum and move the more specialised ones out to separate enums to make it clear that choosing them means you're going somewhat "off piste".

nicktindall:
Fixes #1221

There are two parts to this change.

- The deprecations from and changes to the original `RollCycles` enum
- The new `RollCycle` enums
- Introducing reusable arithmetic
- Use the reusable arithmetic in WeeklyRollCycle
- The deprecation checker
- The new deprecation checks

Which all appear in the [first commit](https://github.com/OpenHFT/Chronicle-Queue/pull/1223/commits/0630a812bc9306f6f465b1a2906c4eee0cbc4cee)

The [second commit](https://github.com/OpenHFT/Chronicle-Queue/pull/1223/commits/fad7dbb2f3cb90616ea801a7c7a2ed91c205b335) is a find/replace replacing all references to the deprecated roll cycles with their replacements

Also got rid of remaining use of the deprecated `ChronicleQueueTestBase` and optimised imports because I came across lots of them in my find/replace-ing


src/main/java/net/openhft/chronicle/queue/rollcycles/LargeRollCycles.java
```diff
@@ -0,0 +1,86 @@
+package net.openhft.chronicle.queue.rollcycles;
+
+import net.openhft.chronicle.core.Maths;
+import net.openhft.chronicle.queue.RollCycle;
+import net.openhft.chronicle.queue.RollCycles;
+
+/**
+ * These are used to minimise rolls but do create very large files, possibly too large.
+ */
+public enum LargeRollCycles implements RollCycle {
```
nicktindall, line 8: Happy to bikeshed the names for these, I honestly don't know what they should be called

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
```diff
@@ -39,7 +38,6 @@
 import java.io.*;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
-import java.util.function.ToLongFunction;
```
JerryShea, line 42: lots of cleanup

pom.xml
```diff
@@ -236,8 +237,7 @@
                         <configuration>
                             <referenceVersion>5.24ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <!-- Reduce the API surface of ChronicleReader -->
-                            <binaryCompatibilityPercentageRequired>100</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>99.9</binaryCompatibilityPercentageRequired>
```
nicktindall, line 239: The return type of `RollCycles.all()` changed from `Iterable<RollCycles>` to `Iterable<RollCycle>`

src/main/java/net/openhft/chronicle/queue/RollCycles.java
```diff
@@ -68,107 +71,164 @@ public enum RollCycles implements RollCycle {
...
+
+    static {
+        List<RollCycle> allCycles = new ArrayList<>();
+        // We can add #values() again once the deprecated ones are gone
+        allCycles.addAll(Arrays.stream(RollCycles.values()).filter(rc -> !RollCycleDeprecationChecker.isDeprecated(rc)).collect(Collectors.toList()));
+        allCycles.addAll(Arrays.asList(LegacyRollCycles.values()));
+        allCycles.addAll(Arrays.asList(LargeRollCycles.values()));
+        allCycles.addAll(Arrays.asList(SparseRollCycles.values()));
+        allCycles.addAll(Arrays.asList(TestRollCycles.values()));
+        VALUES = Collections.unmodifiableList(allCycles);
```
nicktindall, line 218: This is annoying, but now `VALUES` come from multiple enums

src/test/java/net/openhft/chronicle/queue/RollCyclesDistinctnessTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+
+public class RollCyclesDistinctnessTest {
+    
+    @Test
+    public void definedRollCycleFormatsAreDistinct() {
```
nicktindall, line 13: Add a test to ensure the format string is distinct for all our enumerated RollCycles (we use it to determine which roll cycle a queue was created with, in tests at least)

src/main/java/net/openhft/chronicle/queue/rollcycles/RollCycleArithmetic.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.openhft.chronicle.core.Maths;
+import net.openhft.chronicle.queue.RollCycles;
+
+import static net.openhft.chronicle.queue.RollCycle.MAX_INDEX_COUNT;
+
+/**
+ * Reusable optimised arithmetic for converting between cycles/indices/sequenceNumbers
+ */
+public final class RollCycleArithmetic {
```
nicktindall, line 10: This is how we make the arithmetic reusable, it should get in-lined trivially because it's a final class?

JerryShea (approved):

JerryShea:
PR is great but we should probably amend https://github.com/OpenHFT/Chronicle-Queue/blob/ea/docs/FAQ.adoc#how-many-messages-can-be-stored-in-a-chronicle-queue too. Sorry for scope-creep

nicktindall:
> PR is great but we should probably amend https://github.com/OpenHFT/Chronicle-Queue/blob/ea/docs/FAQ.adoc#how-many-messages-can-be-stored-in-a-chronicle-queue too. Sorry for scope-creep

Done this now, and added some code to regenerate the table if/when it changes

PR #1229, opened 2022-10-25
Documentation: use uppercase README like most other Chronicle projects.

JanStureNielsen:


peter-lawrey:
Thanks Jan, and apologies that this sat for so long. The only file renamed by this PR, `docs/draft-docs/readme.adoc`, was removed in `75833a366` during the documentation restructure, while the top-level `README.adoc` has used the uppercase name since 2016. There is therefore nothing left for this PR to change, so I’m closing it as obsolete.

PR #1238, opened 2022-11-01
benchmark for method reader skip

JerryShea:


src/test/java/net/openhft/chronicle/queue/bench/MethodReaderSkipBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+    private MethodReader reader;
+    private int counter = 0;
+
+    public static void main(String[] args) {
+        iterations = Integer.getInteger("benchmarkIterations", 100_000);
+        System.out.println("Iterations: " + iterations);
+        int throughput = Integer.getInteger("benchmarkThroughput", 20_000);
+        System.out.println("Throughput: " + throughput);
+
+        // disable as otherwise single GC event skews results heavily
```
nicktindall: This comment seems orphaned,  I think I recall it being about coordinated omission

JerryShea: good point. Removed. The joy of copy/paste!

nicktindall (approved):

PR #1239, opened 2022-11-03
Fix #https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterpris…

foroughgoudarzi:
…e/issues/370. Add CQE-features.


JerryShea (changes requested):
I don't think there is any point making this change - don't we already have a section on the differences between Queue and QE?

foroughgoudarzi:
> I don't think there is any point making this change - don't we already have a section on the differences between Queue and QE?

The queue docs are being revised currently and I do not know what will be in that (and because of this I delayed this change)
At the moment we only have this about CQE in antora , so I added the features to this section:

Screenshot from 2022-11-04 07-19-46




I thought this was what you want but I think the best place for this list of features is Readme.adoc of CQE. Please let me know if you want to add this to CQE readme for now and later when the queue antora is ready we can make sure it refers to these features.

PR #1245, opened 2022-11-14
Fixed implementation selection when loading config from metadata

### LINKED ISSUE #1244: SingleChronicleQueueBuilder should choose the implementation after config options are loaded from existing metadata
Exisitng `metadata.cq4t` may have some enterprise-only features configured.
If we initialize a queue with `SingleChronicleQueueBuilder.binary(path).build()`, relying on reloading the config from disk, we may choose an incorrect queue implementation.

glukos:
See #1244 
Test (requires QE classes) is in https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/380


JerryShea (approved):

JerryShea:
Suggest we backport this fix to 22 and 23

glukos:
Agreed.
Also, thanks for a quick review!

PR #1249, opened 2022-11-16
Queue enterprise348

alamar:


src/test/java/net/openhft/chronicle/queue/ChronicleQueueIndexTest.java
```diff
@@ -45,13 +45,6 @@
 
 public class ChronicleQueueIndexTest extends QueueTestCommon {
 
-    @Test
-    public void checkTheEOFisWrittenToPreQueueFile() {
```
alamar, line 49: Writing to two roll cycles in the future, no reason why any EOFs will be ever written

src/test/java/net/openhft/chronicle/queue/CreateAtIndexTest.java
```diff
@@ -55,7 +55,7 @@ public class CreateAtIndexTest extends QueueTestCommon {
             String before = queue.dump();
             appender.writeBytes(0x421d00000000L, HELLO_WORLD);
             String after = queue.dump();
-            assertEquals(before, after);
```
alamar: Not sure what happens here, some internal metric gets updated.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -441,14 +438,13 @@ public void normaliseEOFs() {
         }
     }
 
-    private void normaliseEOFs0() {
-        int last = queue.lastCycle();
```
alamar, line 445: 'last' is some random pretouched segment, no reason to loop towards it.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -91,16 +91,18 @@ class StoreAppender extends AbstractCloseable
 
         try {
             queue.cleanupStoreFilesWithNoData();
+            // Avoid normalizing last existing segment which may be appended later on via queue replication
+            //  even if it is in the past
+            int lastCycle = queue.lastCycle();
+            cycle = Math.min(queue.cycle(), lastCycle >= 0 ? lastCycle : queue.cycle());
```
nicktindall: should we try and use a name other than lastCycle for this variable, considering that's also the name of a field in the class?

nicktindall: perhaps even put these two lines into a method with a name that describes what it's doing?

alamar: It does seem suspicious and I will double recheck it.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -91,16 +90,20 @@ class StoreAppender extends AbstractCloseable
 
         try {
             queue.cleanupStoreFilesWithNoData();
+            int lastExistingCycle = queue.lastCycle();
+            // lastCycle() can be negative (no files on disk) or a cycle number of last existing cycle file
+            // Initial cycle should be assigned to last existing cycle file unless it is in the future
+            // If last existing file is in the past, it's still possible it will be replicated/backfilled so no EOF
+            cycle = Math.min(queue.cycle(), lastExistingCycle >= 0 ? lastExistingCycle : queue.cycle());
```
nicktindall: Minor, but it feels like this would be more readable if it were
```java
cycle = lastExistingCycle >= 0 ? Math.min(lastExistingCycle, queue.cycle()) : queue.cycle();
```
Just to make it clear that if `lastExistingCycle < 0` we don't need to do the `Math.min()`

Or even flip the ternary, e.g.
```java
cycle = lastExistingCycle < 0 ? queue.cycle() ? Math.min(lastExistingCycle, queue.cycle());
```

just move it outside the call to `Math.min`

nicktindall: Seems silly to do have a path that does `Math.min(queue.cycle(), queue.cycle())`, unless I've missed something and that's not as nonsensical as it looks.

nicktindall (commented):
This is good @alamar and I approve, but I think we will keep having problems like this as long as pretouched cycles are visible to all this code. I think if we want to make this maintainable, the cycles created by the pretoucher need to be invisible to the code performing all this cycle arithmetic. i.e. it needs to be done transparently in a layer below this stuff. Because any time someone uses `queue.lastCycle()` they are potentially getting a pretouched nothing-cycle. Concerningly that's also a public method on SingleChronicleQueue, so we have a public API giving out bogus info when the pretoucher is in use. It's just counter-intuitive and will continue to be a source of problems until it's fixed.

alamar:
@nicktindall This sounds like a change which should be made after extensive discussion and perhaps in x.25

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
alamar: I've simplified that condition

alamar:
Please re-review @nicktindall @JerryShea

nicktindall (approved):

Issue #1163, opened 2022-09-02
Fix SingleChronicleQueueTest#mappedSegmentsShouldBeUnmappedAsCycleRolls


JerryShea:
This was un-ignored prematurely in #1152

This test should be reviewed in conjunction with other tests related to #1158

JerryShea:
@alamar FYI I ran this test locally and it worked, so I un-ignored it. Unfortunately it was flaky in TeamCity so I re-ignored it; I'm afraid I didn't have the time to investigate.

hft-team-city:
Released in [Chronicle-Queue-5.24ea5](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.24ea5), [BOM-2.24ea17](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24ea17)

PR #1252, opened 2022-11-24
Un-ignore test #1163

alamar:


alamar:
@JerryShea I ran it ~50 times locally and on TC, no failures. I suggest we un-ignore it and observe its behavior.

JerryShea (approved):

PR #1256, opened 2022-11-29
Correct approach of choosing current cycle and EOFing the rest

alamar:
see ChronicleEnterprise/Chronicle-Queue-Enterprise#383


alamar:
Test is in ChronicleEnterprise/Chronicle-Queue-Enterprise#384

nicktindall (approved):
I don't fully understand this change but it seems reasonable to me

PR #1259, opened 2022-12-07
Add portal supplemental UI

julgus:
I've added a folder of supplemental UI files that can be used when building Antora docs for viewing in the portal. Essentially this removes the main menu and the footer, and moves the search bar to the left side menu.


alamar (approved):
Looks OK but let's ask @foroughgoudarzi

docs/antora/site-portal.yml
```diff
@@ -0,0 +1,29 @@
...
+  title: Chronicle Queue
+  start_page: chronicle-queue:introduction:introduction.adoc
+  # The following URL is only used when building locally. In the future, this should be set to the URL
+  # where the page is really served
+  url: /build
+content:
+  sources:
+    - url: https://github.com/OpenHFT/Chronicle-Queue
+      # Only generate from git tags (and not branches)
+      branches: [develop]
```
foroughgoudarzi, line 11: are we moving documentation to the develop branch?

julgus: I have not intended to make any changes regarding what branches we build from, just copied site.yml and edited line 17 which refers to the supplemental build files. If develop is the wrong branch, then we should change that in all the yml build files.

PR #1260, opened 2022-12-08
Backport ExcerptTailer#approximateExcerptsInCycle

nicktindall:
to remove use of SingleChronicleQueue#acquireTailer()

This commit includes changes cherry-picked from https://github.com/OpenHFT/Chronicle-Queue/commit/bfcf600f317898e7bcdbab312604a302ff8411d7 

and also the use of `StoreTailer#exactExcerptsInCycle` in `StoreTailer#tryWindBack`

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -724,6 +724,9 @@ class StoreTailer extends AbstractCloseable
     @NotNull
     @Override
     public final ExcerptTailer toStart() {
+        if (context.isPresent())
+            throw new IllegalStateException("Cannot move tailer to start during document reading");
+
         try {
             return doToStart();
         } catch (MissingStoreFileException e) {
@@ -857,6 +860,9 @@ class StoreTailer extends AbstractCloseable
     @Override
     public ExcerptTailer toEnd() {
         throwExceptionIfClosed();
+
+        if (context.isPresent())
+            throw new IllegalStateException("Cannot move tailer to end during document reading");
 
         if (direction.equals(TailerDirection.BACKWARD)) {
             return callOriginalToEnd();
@@ -1083,7 +1089,7 @@ class StoreTailer extends AbstractCloseable
     }
 
     private boolean tryWindBack(final int cycle) {
-        final long count = queue.exactExcerptsInCycle(cycle);
+        final long count = exactExcerptsInCycle(cycle);
         if (count <= 0)
             return false;
         final RollCycle rollCycle = queue.rollCycle();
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -724,6 +724,9 @@ else if (store != null)
     @NotNull
     @Override
     public final ExcerptTailer toStart() {
+        if (context.isPresent())
+            throw new IllegalStateException("Cannot move tailer to end during document reading");
```
JerryShea: to start?

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -1167,6 +1173,30 @@ public boolean readAfterReplicaAcknowledged() {
...
+
+    @Override
+    public long exactExcerptsInCycle(int cycle) {
+        throwExceptionIfClosed();
+        try {
+            return moveToCycle(cycle) ? store.exactLastSequenceNumber(this) + 1 : -1;
+        } catch (StreamCorruptedException e) {
+            throw new IllegalStateException(e);
+        } finally {
+            releaseStore();
```
JerryShea, line 1196: @alamar let's discuss

alamar:
I thought of this as of distinctive breaking change of 2.24 - do we really need to backport it? It changes API

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
alamar: I no longer remember what it is about, but we already pushed it to 2.24ea - if there is any problem with it, we should probably raise an issue to fix it in 2.24ea first.

nicktindall:
> I thought of this as of distinctive breaking change of 2.24 - do we really need to backport it? It changes API

When you say it changes the API, it does add two methods to the ExcerptTailer interface, and adds the side-effects into `tryWindBack`, but that's a private method. Does any of that really matter? I did leave the existing methods on queue and they still use `acquireTailer`, so anyone calling them should get the same behaviour.

The reason we want to do it is to stop the use of `acquireTailer` by FIX infrastructure, because it causes file handles to be left open on roll cycles.

alamar (approved):
Approved assuming this is straight cherry-pick of existing commit.

nicktindall:
> Approved assuming this is straight cherry-pick of existing commit.

It's a cherry-pick of a subset of the changes from the existing commit. Differences are:
- Didn't remove `SingleChronicleQueueStore#acquire(int, long, boolean, SingleChronicleQueueStore)`
- Applied a non-breaking subset of the changes to `BinarySearch` (add the new tailer-only signature for `findWithinCycle`, but left the old one that takes a queue for backward compatibility, left `search` as-is)
- Left `SingleChronicleQueue#acquireTailer` there
- Deprecated `SingleChronicleQueue#(approximate|exact)ExcerptsInCycle()`, but left the implementations as they were (using `acquireTailer`)

In addition, I changed `StoreTailer#tryWindBack(int)` to use `ExcerptTailer#exactExcerptsInCycle()` instead of `SingleChronicleQueue#exactExcerptsInCycle(int)`, but I also applied this change to develop (and it's been merged)

PR #1270, opened 2022-12-14
fix redirecting links

j4sm1ne96:
fix any broken internal links, use href


foroughgoudarzi (changes requested):
please do not commit site.css and index.htm

PR #1274, opened 2022-12-18
Stablization of ungraceful restarts test

### LINKED ISSUE #1273: Stabilize performance of exact excerpt counting
Caused by #1203 
It seems this check works in disabling way currently:
```
        if (((Byteable) index2indexArr).bytesStore() != null)
            return 0;
```
Leading to having 0 returned every time, and full scan every time.

alamar:
ChronicleEnterprise/Chronicle-Queue-Enterprise#333

closes #1273


alamar:
https://teamcity.chronicle.software/project/Chronicle_ChronicleQueue?projectTab=stats&branch= 
I have run some performance benchmarks but don't see anything definitive
(four on the left - default, four on the right - with this PR)

alamar:
This change makes tailing more robust, as well as fixing condition bug which prevented indexed access and adding extra synchronization before writing excerpt (which fixes the actual issue)

peter-lawrey (approved):

PR #1279, opened 2023-01-05
ChronicleReader support named tailer. Closes #1277

### LINKED ISSUE #1277: ChronicleReader support named tailer
To allow resumption

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -812,7 +812,7 @@ void position0(final long position, final long startOfMessage, Bytes<?> bytes) {
 
     @Override
     public @NotNull StoreAppender disableThreadSafetyCheck(boolean disableThreadSafetyCheck) {
-        super.disableThreadSafetyCheck(disableThreadSafetyCheck);
+        super.singleThreadedCheckDisabled(disableThreadSafetyCheck);
```
JerryShea, line 815: reduce call to deprecated methods

src/main/java/net/openhft/chronicle/queue/reader/ChronicleReader.java
```diff
@@ -98,11 +97,10 @@ public void execute() {
         boolean queueHasBeenModified;
         do {
             try (final ChronicleQueue queue = createQueue();
-                 final ExcerptTailer tailer = queue.createTailer()) {
+                 final ExcerptTailer tailer = queue.createTailer(tailerId);
+                 final ExcerptTailer toEndTailer = queue.createTailer()) {
                 MessageHistory.set(new VanillaMessageHistory());
 
-                tlTailer = ThreadLocal.withInitial(queue::createTailer);
```
JerryShea, line 104: TL was not required - this is all done in 1 thread

rogersimmons (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: We should probably deprecate that method as well (Deprecated annotation is not inherited) I think if the method calling the deprecated method is also deprecated the compiler/sonar/intelliJ warning goes away too

src/test/java/net/openhft/chronicle/queue/internal/reader/ChronicleReaderTest.java
```diff
@@ -92,7 +92,8 @@ public void before() {
         if (OS.isWindows())
             if (!(testName.getMethodName().equals("shouldThrowExceptionIfInputDirectoryDoesNotExist") ||
                     testName.getMethodName().equals("shouldBeAbleToReadFromReadOnlyFile") ||
-                    testName.getMethodName().equals("shouldPrintTimestampsToLocalTime")))
+                    testName.getMethodName().equals("shouldPrintTimestampsToLocalTime") ||
+                    testName.getMethodName().equals("matchLimitThenNamedTailer")))
```
nicktindall, line 97: Do we still get this even though we explicitly set readOnly to false?

nicktindall (approved):
Approved with minor comment & question

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
JerryShea: The Intellij warning does *not* go away unfortunately, and so I didn't bother deprecating it. We haven't done that consistently anywhere else I don't believe

src/test/java/net/openhft/chronicle/queue/internal/reader/ChronicleReaderTest.java
JerryShea: This condition is negated

nicktindall: We could just use `ignoreException` here, it will not fail if the warning isn't present

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: I usually try to when I deprecate anything

nicktindall: It does go away in 2022.3.1

Screenshot from 2023-01-06 11-26-52
Screenshot from 2023-01-06 11-27-00

PR #1292, opened 2023-01-30
backport Pretoucher fix from 2.24

JerryShea:


alamar:
Wasn't the whole point disabling/not recommending pretoucher in 2.23 and before that? I believe that's what we have discussed two months back.

nicktindall:
This doesn't fix the break in CQE

JerryShea:
In x.24 Pretoucher has been moved to QE, but it is still in 23 stable, so we should apply this fix

alamar:
What's the status of this change?

JerryShea:
Closing - no point in doing this

PR #1294, opened 2023-02-05
It is now not possible to write over EOF

JerryShea:
https://github.com/OpenHFT/Chronicle-Wire/issues/618


nicktindall (approved):
Pending maybe testing for a more specific exception if we add it

Issue #1291, opened 2023-01-30
OverlappingFileLockException when headerRecovery results in resizeRafIfTooSmall


nicktindall:
When `net.openhft.chronicle.queue.impl.single.SingleChronicleQueue.StoreSupplier#headerRecovery` executes and it results in a `net.openhft.chronicle.bytes.internal.ChunkedMappedFile#resizeRafIfTooSmall`, you end up attempting to lock the file channel twice. The lock is not re-entrant, so the OverlappingFileLockException is thrown. We might need to implement a re-entrant layer in ChunkedMappedFile or something.

Example stacktrace:
```
Exception in thread "main" java.nio.channels.OverlappingFileLockException
  at sun.nio.ch.SharedFileLockTable.checkList(FileLockTable.java:255)
  at sun.nio.ch.SharedFileLockTable.add(FileLockTable.java:152)
  at sun.nio.ch.FileChannelImpl.lock(FileChannelImpl.java:1062)
  at java.nio.channels.FileChannel.lock(FileChannel.java:1053)
  at net.openhft.chronicle.bytes.internal.ChunkedMappedFile.resizeRafIfTooSmall(ChunkedMappedFile.java:258)
  at net.openhft.chronicle.bytes.internal.ChunkedMappedFile.acquireByteStore(ChunkedMappedFile.java:181)
  at net.openhft.chronicle.bytes.MappedFile.acquireByteStore(MappedFile.java:197)
  at net.openhft.chronicle.bytes.internal.ChunkedMappedBytes.acquireNextByteStore0(ChunkedMappedBytes.java:338)
  at net.openhft.chronicle.bytes.internal.ChunkedMappedBytes.readVolatileInt(ChunkedMappedBytes.java:468)
  at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.headerRecovery(SingleChronicleQueue.java:1089)
  at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue$StoreSupplier.acquire(SingleChronicleQueue.java:1043)
  at net.openhft.chronicle.queue.impl.WireStorePool.acquire(WireStorePool.java:56)
  at net.openhft.chronicle.queue.impl.single.SingleChronicleQueue.storeForCycle(SingleChronicleQueue.java:574)
  at net.openhft.chronicle.queue.impl.single.StoreTailer.cycle(StoreTailer.java:1128)
[...]

PR #1295, opened 2023-02-07
Reproduction and fix for #1291, #1296

nicktindall:
Test that Tailers and Appenders handle an empty roll cycle at the end of the queue gracefully. This appears to happen sometimes when a writing process is kill -9'd

Depends on https://github.com/OpenHFT/Chronicle-Bytes/pull/488

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -18,6 +18,7 @@
 package net.openhft.chronicle.queue.impl.single;
 
 import net.openhft.chronicle.bytes.*;
+import net.openhft.chronicle.bytes.domestic.ReentrantFileLock;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1110,7 +1114,9 @@ private synchronized void headerRecovery(final SingleChronicleQueue that, final
                     }
                 } finally {
                     try {
-                        fileLock.release();
+                        if (fileLock != null) {
```
JerryShea: won't it always be non null?

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
nicktindall: No the earlier null check is inside the try block. I should move it out though.

PR #1299, opened 2023-02-14
Reset position on the roll to prevent replication stopping

alamar:
see ChronicleEnterprise/Chronicle-Queue-Enterprise#413


alamar:
I'm not sure why it was not causing issues on develop, but I guess it makes complete sense to apply it to both develop and 2.23

nicktindall (approved):
:raised_hands:

JerryShea (approved):

PR #1304, opened 2023-02-17
Avoid publishing partially filed out header

alamar:
see ChronicleEnterprise/Chronicle-Queue-Enterprise#415


nicktindall (approved):
This looks good to me, but @peter-k-lawrey has the final say

JerryShea (approved):
This is excellent. Would prefer we get @peter-lawrey tick of approval before merging though

PR #1311, opened 2023-02-28
Store the index we've normalisedEOFs to...

### LINKED ISSUE #1306: Optimise normaliseEOFs
`normaliseEOFs` runs from the first cycle to the (current - 1) cycle writing any missing EOFs. This can take a significant amount of time on a queue with many historic roll cycles present. It shouldn't be necessary if we ensure the previous roll cycle has an EOF prior to writing any entries to a new one.

We will still need to keep the corrective action for queues created by old versions that did not ensure this, but we only need to do that scan once.

-- Details to follow --

nicktindall:
... to avoid re-normalising the queue from the beginning each time. Fixes #1306

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -921,7 +921,7 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
     }
 
     @Nullable
-    protected LongValue tableStoreAcquire(CharSequence key, long index) {
+    protected LongValue tableStoreAcquire(CharSequence key, long defaultValue) {
 
         BytesStore keyBytes = asBytes(key);
         LongValue longValue = metaStoreMap.get(keyBytes);
@@ -929,12 +929,12 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
             synchronized (closers) {
                 longValue = metaStoreMap.get(keyBytes);
                 if (longValue == null) {
-                    longValue = metaStore.acquireValueFor(key, index);
+                    longValue = metaStore.acquireValueFor(key, defaultValue);
                     int length = key.length();
                     HeapBytesStore<byte[]> key2 = HeapBytesStore.wrap(new byte[length]);
                     key2.write(0, keyBytes, 0, length);
                     metaStoreMap.put(key2, longValue);
-                    return null;
+                    return longValue;
                 }
             }
         }
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
@@ -1,0 +1,96 @@
+package net.openhft.chronicle.queue.impl.single;
+
+import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.io.IOTools;
+import net.openhft.chronicle.core.onoes.ExceptionKey;
+import net.openhft.chronicle.core.onoes.LogLevel;
+import net.openhft.chronicle.core.time.SetTimeProvider;
+import net.openhft.chronicle.core.time.TimeProvider;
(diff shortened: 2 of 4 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -921,20 +921,20 @@ public void tableStorePut(CharSequence key, long index) {
...
                 longValue = metaStoreMap.get(keyBytes);
                 if (longValue == null) {
-                    longValue = metaStore.acquireValueFor(key, index);
+                    longValue = metaStore.acquireValueFor(key, defaultValue);
                     int length = key.length();
                     HeapBytesStore<byte[]> key2 = HeapBytesStore.wrap(new byte[length]);
                     key2.write(0, keyBytes, 0, length);
                     metaStoreMap.put(key2, longValue);
-                    return null;
+                    return longValue;
```
nicktindall, line 937: I don't know why this was returning null?

JerryShea (approved):
nice

src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Before
+    @Override
+    public void recordExceptions() {
+        super.recordExceptions();
+        exceptionMap = Jvm.recordExceptions(true);
+        exceptionTracker = ExceptionTracker.create(ExceptionKey::message, ExceptionKey::throwable, Jvm::resetExceptionHandlers, exceptionMap,
+                (key) -> key.level != LogLevel.DEBUG && key.level != LogLevel.PERF,
+                (key) -> key.level() + " " + key.clazz().getSimpleName() + " " + key.message());
+        ignoreException(ex -> true, "Ignore everything");
```
nicktindall, line 50: This is a bit naughty. We probably should add a way to ask for debug exceptions and iterate over the exceptions in order.

rogersimmons (commented):
Is this feature needed?
Isn't it enough to walk backwards from the most recent, writing EOFs where missing, then stopping at the first file with an EOF already present?

nicktindall:
> Is this feature needed?
> Isn't it enough to walk backwards from the most recent, writing EOFs where missing, then stopping at the first file with an EOF already present?
> 

Apparently not. We did prototype that and it appeared to work but @alamar [made the point](https://github.com/OpenHFT/Chronicle-Queue/issues/1306#issuecomment-1439896535) that there can be numerous pre-touched (EOF-less) cycles present when earlyAcquireNextCycle is turned on.

I would much prefer to do this without adding state, but until we've made the early acquired cycles distinguishable from actual cycles I fear that's not a possibility. This should be considered a stop gap until we can do that or similar.

rogersimmons:
ok, but isn't this the problem which needs to be fixed then (rather than adding more state to tolerate it)?
"there can be numerous pre-touched (EOF-less) cycles present when earlyAcquireNextCycle is turned on"

nicktindall:
> ok, but isn't this the problem which needs to be fixed then (rather than adding more state to tolerate it)?
> "there can be numerous pre-touched (EOF-less) cycles present when earlyAcquireNextCycle is turned on" 

Yes, agreed

rogersimmons (changes requested):
Could you add an issue (tagged as high priority) to address the root cause and link it back to this ticket pls.
Also wrap all changes on this ticket with an explicit comment so they can easily be removed in due course. (The hash/annotate alone won't be enough as may not survive long).

Will then approve 
Thanks

nicktindall:
> Could you add an issue (tagged as high priority) to address the root cause and link it back to this ticket pls. Also wrap all changes on this ticket with an explicit comment so they can easily be removed in due course. (The hash/annotate alone won't be enough as may not survive long).
> 
> Will then approve Thanks

I created https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/issues/423 which should remove the need for normaliseEOFs altogether. This code will all be deleted naturally then because it's primarily to do with normaliseEOFs.

rogersimmons (approved):
Thanks Nick. Have approved

PR #1314, opened 2023-03-07
Rename and recreate un-initialized cycle file instead of overwriting

### LINKED ISSUE #1313: Instead of recovering header, rename file then create a new one
Make sure concurrent creation is handled as well

see also https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/issues/415

alamar:
closes #1313

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -18,7 +18,6 @@
 package net.openhft.chronicle.queue.impl.single;
 
 import net.openhft.chronicle.bytes.*;
-import net.openhft.chronicle.bytes.domestic.ReentrantFileLock;
 import net.openhft.chronicle.bytes.internal.HeapBytesStore;
 import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.OS;
@@ -27,6 +26,7 @@ import net.openhft.chronicle.core.annotation.PackageLocal;
 import net.openhft.chronicle.core.announcer.Announcer;
 import net.openhft.chronicle.core.io.AbstractCloseable;
 import net.openhft.chronicle.core.io.Closeable;
+import net.openhft.chronicle.core.io.ReferenceOwner;
 import net.openhft.chronicle.core.threads.CleaningThreadLocal;
 import net.openhft.chronicle.core.threads.EventLoop;
 import net.openhft.chronicle.core.threads.InterruptedRuntimeException;
@@ -45,9 +45,6 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.*;
-import java.nio.channels.ClosedChannelException;
-import java.nio.channels.FileLock;
-import java.nio.channels.NonWritableChannelException;
 import java.security.SecureRandom;
 import java.text.ParseException;
 import java.time.ZoneId;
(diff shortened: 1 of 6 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -70,6 +67,7 @@
 public class SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue {
 
     public static final String SUFFIX = ".cq4";
+    public static final String DISCARD_FILE_SUFFIX = ".discard";
```
rogersimmons, line 69: I would suggest a sequence or timestamp-based decoration is a better/more flexible option

rogersimmons (changes requested):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
alamar: Timestamp-based decoraton means we may copy a single cycle an arbitrary number of times, which in case of high performance system might as well be tens of thousands times which will exhaust inodes and-or current directory files limit. That's why I advise we use predictable discard name and avoid this issue entirely. If it's not successful after first try it's probably too dangerous to continue - what do you think @rogersimmons ?

rogersimmons: Yes, I agree with the concern, and it is something we should be in a position to mitigate.
Having a sequence/time-based decoration helps rather than impedes that, and at the same time gives us more control options in future if needed

alamar: I don't know how we could mitigate this *right now*, whereas the current solution is indeed limited but is predictable and completely understood, and we can reason about its behavior.

I suggest that maybe we add time-based decoration together with the change that protects about creating the same cycle file unbounded number of times.

src/test/java/net/openhft/chronicle/queue/impl/single/EmptyRollCycleTest.java
```diff
@@ -108,7 +106,7 @@ public void appropriateExceptionIsThrownWhenLockCannotBeAcquiredForRecovery() th
                         assertTrue(readingDocument.isPresent());
                     }
                 }
-                assertThrows(IOException.class, tailer::readingDocument);
+                assertEquals(false, tailer.readingDocument().isPresent());
```
nicktindall, line 109: perhaps rename the test also (now an exception is not thrown)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1079,54 +1092,15 @@ public SingleChronicleQueueStore acquire(int cycle, boolean createIfAbsent) {
...
-                    }
-
-                    if (!wire.writeFirstHeader()) {
-                        Jvm.warn().on(getClass(), "failed to recover.");
-                        throw new StreamCorruptedException("failed to recover.˚");
-                    }
+        // Rename un-acquirable segment file to make sure no data is lost to try and recreate the segment
+        private boolean backupCycleFile(File cycleFile) {
+            File cycleFileDiscard = new File(cycleFile.getParentFile(), cycleFile.getName() + DISCARD_FILE_SUFFIX);
+            boolean success = cycleFile.renameTo(cycleFileDiscard);
```
nicktindall, line 1099: Is it possible that multiple processes get here and think they moved the existing file (i.e is the move atomic), and would that cause chaos later if they both attempted to create the replacement file? I notice [Files.move](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#move-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-) allows you to specify an atomic move (and define the behaviour when an existing file is there), not sure if that would be more appropriate here.

nicktindall (approved):
Approved pending resolution of the discussion around timestamped backups

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
alamar: Only one of these process succeeds and this one gets to recreate the file from scratch - other ones retry opening this file without createIfAbsent flag. I find the behavior of File.renameTo() more consistent than the described atomic move, but when we rename file in the current directory it should almost always be atomic yes or no operation.

rogersimmons (approved):

PR #1316, opened 2023-03-16
Correct calculation of last index when it is not in the latest segment

alamar:
see #ChronicleEnterprise/Chronicle-Queue-Enterprise#443

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/WireStoreSupplier.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/WireStoreSupplier.java
@@ -26,8 +26,14 @@ import java.text.ParseException;
 import java.util.NavigableSet;
 
 public interface WireStoreSupplier {
+    enum CreateStrategy {
+        CREATE, // Always create file
+        REINITIALIZE_EXISTING, // Do not create new file but reinitialize file if header is not ready - for normalizeEOF
+        READ_ONLY
+    }
+
     @Nullable
-    SingleChronicleQueueStore acquire(int cycle, boolean createIfAbsent);
+    SingleChronicleQueueStore acquire(int cycle, CreateStrategy createStrategy);
 
     /**
      * the next available cycle, no cycle will be created by this method, typically used by a
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -776,11 +776,13 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
     public long lastIndex() {
         // This is a slow implementation that gets a Tailer/DocumentContext to find the last index
         try (final ExcerptTailer tailer = createTailer().direction(BACKWARD).toEnd()) {
-            try (final DocumentContext documentContext = tailer.readingDocument()) {
-                if (documentContext.isPresent()) {
-                    return documentContext.index();
-                } else {
-                    return -1;
+            while (true) {
+                try (final DocumentContext documentContext = tailer.readingDocument()) {
+                    if (documentContext.isPresent() && !documentContext.isMetaData()) {
+                        return documentContext.index();
+                    } else {
+                        return -1;
+                    }
                 }
             }
         }
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -569,7 +569,9 @@ public class SingleChronicleQueue extends AbstractCloseable implements RollingCh
     @Nullable
     @Override
     public final SingleChronicleQueueStore storeForCycle(int cycle, final long epoch, boolean createIfAbsent, SingleChronicleQueueStore oldStore) {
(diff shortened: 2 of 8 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -776,11 +776,13 @@ public long firstIndex() {
...
-                    return documentContext.index();
-                } else {
-                    return -1;
+            while (true) {
+                try (final DocumentContext documentContext = tailer.readingDocument()) {
+                    if (documentContext.isPresent() && !documentContext.isMetaData()) {
+                        return documentContext.index();
+                    } else {
+                        return -1;
+                    }
```
nicktindall, line 788: is the while (true) redundant here? it seems regardless of which branch you take you return something on the first iteration?

nicktindall: I'm guessing in the event it's metadata we want to keep scanning backwards?

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -795,8 +794,11 @@ private long approximateLastCycle2(int lastCycle) throws StreamCorruptedExceptio
                     // try again.
                 }
             }
+            this.setCycle(lastCycle);
             return rollCycle.toIndex(lastCycle, 0L);
```
nicktindall: If we get here, lastCycle == prevCycle, correct? would it not be clearer to 
```java
this.setCycle(prevCycle);
return rollCycle.toIndex(prevCycle, 0L);
```

Also should `prevCycle` be called `firstCycle` to make it clearer what's going on here?

nicktindall (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
JerryShea: and we should have a timeout rather than loop forever?

src/main/java/net/openhft/chronicle/queue/impl/WireStoreSupplier.java
```diff
@@ -26,8 +26,14 @@
...
 public interface WireStoreSupplier {
+    enum CreateStrategy {
+        CREATE, // Always create file
+        REINITIALIZE_EXISTING, // Do not create new file but reinitialize file if header is not ready - for normalizeEOF
+        READ_ONLY
+    }
+
     @Nullable
-    SingleChronicleQueueStore acquire(int cycle, boolean createIfAbsent);
+    SingleChronicleQueueStore acquire(int cycle, CreateStrategy createIfAbsent);
```
nicktindall: Rename parameter name to createStrategy (and in implementation)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
nicktindall: It won't loop forever because it's traversing the queue backwards, it'll hit the first cycle then exit.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -975,7 +979,7 @@ private StoreSupplier() {
 
         @SuppressWarnings("resource")
         @Override
-        public SingleChronicleQueueStore acquire(int cycle, boolean createIfAbsent) {
+        public SingleChronicleQueueStore acquire(int cycle, CreateStrategy createIfAbsent) {
```
nicktindall: Parameter name

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -297,10 +298,10 @@ public boolean recordHistory() {
 
     void setCycle(int cycle) {
         if (cycle != this.cycle)
-            setCycle2(cycle, true);
+            setCycle2(cycle, WireStoreSupplier.CreateStrategy.CREATE);
     }
 
-    private void setCycle2(final int cycle, final boolean createIfAbsent) {
+    private void setCycle2(final int cycle, final WireStoreSupplier.CreateStrategy createIfAbsent) {
```
nicktindall: Parameter name

nicktindall (changes requested):

alamar:
@peter-lawrey @nicktindall @JerryShea please re-review - this change finally stabilizes the UngracefulRestartsTest.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
nicktindall: This one's not done

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: And this one :) I think you fixed the two I didn't see

alamar:
@nicktindall I hope I've fixed the remaining ones

nicktindall (approved):

PR #1320, opened 2023-03-30
Make SingleChronicleQueueStore handle the case where we're closing a partially initialised instance

nicktindall:
If the constructor throws, close gets called in the constructor. That can mean the `mappedBytes` field can be null.

Revealed due to fix for https://github.com/OpenHFT/Chronicle-Core/issues/499


JerryShea (approved):
nice one

PR #1325, opened 2023-04-17
Unlock queue lock in event of writeAfterEOFException. Fixes #1324

### LINKED ISSUE #1324: WriteAfterEOFException thrown by writingDocument() leaves queue lock locked
When an attempt is made to write to an EOF'd roll cycle, `WriteAfterEOFException` is thrown and it leaves the write lock locked. While unlikely to occur, this shouldn't leave the queue in a non-writable state.

nicktindall:
Fixes #1324

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -311,7 +311,7 @@ class StoreAppender extends AbstractCloseable
 
         SingleChronicleQueueStore oldStore = this.store;
 
-        SingleChronicleQueueStore newStore = storePool.acquire(cycle,  createStrategy, oldStore);
+        SingleChronicleQueueStore newStore = storePool.acquire(cycle, createStrategy, oldStore);
 
         if (newStore != oldStore) {
             this.store = newStore;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -427,22 +427,28 @@ private StoreAppender.StoreAppenderContext prepareAndReturnWriteContext(boolean
...
-            // sets the writeLimit based on the safeLength
-            openContext(metaData, safeLength);
+                // sets the writeLimit based on the safeLength
+                openContext(metaData, safeLength);
 
-            // Move readPosition to the start of the context. i.e. readRemaining() == 0
-            wire.bytes().readPosition(wire.bytes().writePosition());
+                // Move readPosition to the start of the context. i.e. readRemaining() == 0
+                wire.bytes().readPosition(wire.bytes().writePosition());
+            } catch (WriteAfterEOFException e) {
```
nicktindall: I wonder if it's safe to generalise this to RuntimeException. It feels a bit less hacky to only unlock for a specific failure that we know the consequences of.

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: I had to change it to `RuntimeException` because the Java 11 build failed mysteriously when I caught `WriteAfterEOFException`

PR #1335, opened 2023-05-09
don't refresh file listing if !force

JerryShea:


alamar:
almost all tests has failed @JerryShea - I wonder if it is caused by this change, if our tests do not pass perhaps we should remove force flag at all (from both table and file system stores, also perhaps add better coverage of table store)

I find the check in two places highly suspicious:

```
        long now = System.currentTimeMillis();
        if (now <= directoryListing.lastRefreshTimeMS()) { // lastRefreshTimeMS is also a currentTimeMillis()
            return;
        }
```

Perhaps it should be turned into assert, though in multi-thread program it may occassionally fail (but then so what?)

But back to your original change, we have two `refresh(false)` in `SingleChronicleQueue` and these no longer have any meaning anymore - do they? What's the point of having these calls?

alamar:
See #1355

JerryShea:
Only one test failed but the approach is not safe. This has been superseded by #1363

PR #1348, opened 2023-05-19
Filtered subscription

peter-lawrey:


RobAustin:
what about subscribeSourceId ? could you add this as well

PR #1357, opened 2023-06-06
Issue1355

alamar:


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -767,6 +767,7 @@ public final int cycle(TimeProvider timeProvider) {
     public long firstIndex() {
         // TODO - as discussed, peter is going find another way to do this as this solution
```
JerryShea, line 768: @peter-lawrey can you review this comment please?

PR #1358, opened 2023-06-07
Feature/more perf tests

JerryShea:


src/test/java/net/openhft/chronicle/queue/impl/single/stress/RollCycleMultiThreadStressSharedWriterQueueTest.java
```diff
@@ -32,6 +32,6 @@ public void stress() throws Exception {
     }
 
     public static void main(String[] args) throws Exception {
-        new RollCycleMultiThreadStressSharedWriterQueueTest().stress();
+        new RollCycleMultiThreadStressSharedWriterQueueTest().main();
```
nicktindall: `main` would be a static method wouldn't it? Can we just do:

```
RollCycleMultiThreadStressSharedWriterQueueTest.main(args);
```

Otherwise it looks like we're creating an instance unnecessarily

nicktindall: Oh never mind, that's very confusing though... perhaps the old naming was less confusing?

nicktindall: Or even `run()`?

pom.xml
```diff
@@ -609,7 +621,30 @@
...
-
+                            <execution>
+                                <id>RollCycleMultiThreadStressTest</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <executable>${java.home}/bin/java</executable>
+                                    <commandlineArgs>${jvm.requiredArgs} -Djvm.resource.tracing=false -DtestTime=30 -classpath %classpath net.openhft.chronicle.queue.impl.single.stress.RollCycleMultiThreadStressTest
```
alamar, line 633: Will it work like a benchmark e.g. expose data for TC to visualise?

src/test/java/net/openhft/chronicle/queue/impl/single/stress/RollCycleMultiThreadStressSharedWriterQueueTest.java
JerryShea: Yep OK fair enough I will change to run

pom.xml
JerryShea: Yes - see [stats graph](https://teamcity.chronicle.software/project/Chronicle_ChronicleQueue?projectTab=stats&branch=feature%2Fmore-perf-tests&mode=builds)

nicktindall (approved):

PR #1363, opened 2023-06-13
Allow a readOnly queue to use TableDirectoryListing. Closes #1355

### LINKED ISSUE #1355: Check if directory listing checks may be skipped
Right now we skip table store listings if force flag is off, but we do not do the same for file system directory listings - need to check if that is possible.

See #1335

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
```diff
@@ -64,16 +75,24 @@ final class TableDirectoryListing extends AbstractCloseable implements Directory
...
-            return this;
-        });
+        final Runnable initLongValues = () -> {
+            maxCycleValue = tableStore.acquireValueFor(HIGHEST_CREATED_CYCLE);
+            minCycleValue = tableStore.acquireValueFor(LOWEST_CREATED_CYCLE);
+            modCount = tableStore.acquireValueFor(MOD_COUNT);
+        };
+
+        if (readOnly) {
+            initLongValues.run();
```
JerryShea: if we are racing with a writeable TableDirectoryListing then it is possible this may fail

nicktindall: We'll just need a version of `TableStore#acquireValue` that doesn't create it if it doesn't exist (e.g. `TableStore#acquireValueIfExists`?). Looking at the code that doesn't seem that tricky?

nicktindall: Ah I see you throw now if it's read-only and doesn't exist. Nice.

src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
```diff
@@ -41,21 +41,32 @@ final class TableDirectoryListing extends AbstractCloseable implements Directory
     private final TableStore<?> tableStore;
     private final Path queuePath;
     private final ToIntFunction<String> fileNameToCycleFunction;
+    private final boolean readOnly;
     private volatile LongValue maxCycleValue;
     private volatile LongValue minCycleValue;
     private volatile LongValue modCount;
     private long lastRefreshTimeMS = 0;
 
+    @Deprecated(/* to be removed in x.25 */)
```
nicktindall: Propose we create a separate `ReadOnlyTableDirectoryListing` instead of adding all the extra branching in. They share almost no logic, and the `readOnly` flag only adds undue complexity to the writable version.

JerryShea: Done, although with inheritance ;)

alamar (approved):
So it *would* pick up newly created segments even in read-only mode? Approve.

PR #1364, opened 2023-06-14
TableDirectoryListingReadOnly protect itself from concurrent creation vs TableDirectoryListing. #1355

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListingReadOnly.java
```diff
@@ -38,7 +39,18 @@ protected void checkReadOnly(@NotNull TableStore<?> tableStore) {
...
+        // it is possible if r/o queue created at same time as r/w queue for longValues to be only half-written
+        final long timeoutMillis = System.currentTimeMillis() + 500;
+        while (true) {
+            try {
+                initLongValues();
+                break;
+            } catch (Exception e) {
+                if (System.currentTimeMillis() > timeoutMillis)
+                    throw e;
+                Jvm.pause(1);
```
nicktindall, line 51: If there's an error here I think we need to close any of the LongValues that were successfully acquired first, or we'll be leaving them open when they're re-assigned on the next iteration

nicktindall: Or if read-only were a distinct implementation, it's initLongValues could only try and acquire values it hasn't already acquired ;)

src/test/java/net/openhft/chronicle/queue/DirectoryUtils.java
```diff
@@ -31,6 +31,7 @@ public static File tempDir(String name) {
         String replacedName = name.replaceAll("[\\[\\]\\s]+", "_").replace(':', '_');
         final File tmpDir = new File(OS.getTarget(), replacedName + "-" + uniqueId());
         DeleteStatic.INSTANCE.add(tmpDir);
+        tmpDir.deleteOnExit();
```
nicktindall, line 34: It looks as though the previous line _should_ take care of this already?

JerryShea: just best practice to set these things as deleteOnExit as shutdown hooks may not get run

src/test/java/net/openhft/chronicle/queue/ReadWriteTest.java
```diff
@@ -140,13 +128,18 @@ public void testProceedWhenMetadataFileInitialized() throws IOException {
             }
         }).start();
 
-        // This should succeed after 200ms
+        // the below can happen if the race mitigation code in TableDirectoryListingReadOnly.init is exercised
+        // as a LongValue gets created before it can be assigned to a reference and be available to be closed
+        ignoreException("Discarded without closing");
```
nicktindall, line 133: Ah yeah this, we can fix this with my comment above I believe

src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListingReadOnly.java
JerryShea: Problem is, the LongValues won't get assigned to a variable if an exception is thrown, so closing them will only close the ones that were successfully acquired

nicktindall (changes requested):
Clean up before retry

nicktindall (approved):

src/test/java/net/openhft/chronicle/queue/ReadWriteTest.java
JerryShea: No, because an exception can be thrown after the LongValue is created and before it is returned

PR #1368, opened 2023-06-14
Handle deleted roll cycles more gracefully

nicktindall:
At client they were encountering MissingFileStoreExceptions after they did a cleanup of "old" roll cycles.

The reason is they were deleting some cycles that were currently being read by a Tailer.

There were a number of places where we didn't handle roll cycles that went missing so gracefully. This PR addresses some of those, but there remains some more (when the first roll cycle is deleted).

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -532,7 +532,6 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
      * While we are doing this we lock the queue so that new appenders can not be created.
      * <p>
      * Queue locks have no impact if you are not using queue replication because the are implemented as a no-op.
-     *
      * @deprecated this is being removed in x.25 with no replacement
      */
     @Deprecated(/* To be removed in x.25 */)
@@ -1222,42 +1221,32 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
             NavigableMap<Long, File> tree = cycleTree(false);
             final File currentCycleFile = dateCache.resourceFor(currentCycle).path;
 
+            // confirm the current cycle is in the min/max range, delay and refresh
+            // a few times if not as this suggests files have been deleted
             directoryListing.refresh(false);
             if (currentCycle > directoryListing.getMaxCreatedCycle() ||
                     currentCycle < directoryListing.getMinCreatedCycle()) {
-                boolean fileFound = false;
                 for (int i = 0; i < 20; i++) {
                     Jvm.pause(10);
                     directoryListing.refresh(i > 1);
-                    fileFound = (
-                            currentCycle <= directoryListing.getMaxCreatedCycle() &&
-                                    currentCycle >= directoryListing.getMinCreatedCycle());
-                    if (fileFound) {
+                    if (currentCycle <= directoryListing.getMaxCreatedCycle() &&
+                            currentCycle >= directoryListing.getMinCreatedCycle()) {
                         break;
                     }
                 }
-                fileFound |= currentCycleFile.exists();
-
-                if (!fileFound) {
-                    directoryListing.refresh(true);
-                    throw new MissingStoreFileException(
-                            String.format("Expected file to exist for cycle: %d, file: %s.%nminCycle: %d, maxCycle: %d%n" +
-                                            "Available files: %s",
-                                    currentCycle, currentCycleFile,
-                                    directoryListing.getMinCreatedCycle(), directoryListing.getMaxCreatedCycle(),
-                                    Arrays.toString(path.list((d, n) -> n.endsWith(SingleChronicleQueue.SUFFIX)))));
-                }
             }
 
+            // check that the current cycle is in the tree, do a hard refresh and retry if not
             Long key = dateCache.toLong(currentCycleFile);
             File file = tree.get(key);
-            // already checked that the file should be on-disk, so if it is null, call cycleTree again with force
             if (file == null) {
                 tree = cycleTree(true);
                 file = tree.get(key);
             }
+
+            // The current cycle is no longer on disk, log an error
             if (file == null) {
-                throw new MissingStoreFileException("missing currentCycle, file=" + currentCycleFile);
+                Jvm.error().on(SingleChronicleQueue.class, "The current cycle seems to have been deleted from under the queue, scanning to find the next remaining cycle, currentCycle=" + currentCycleFile);
             }
 
             switch (direction) {
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -536,11 +536,17 @@ class StoreTailer extends AbstractCloseable
             nextIndex = nextIndexWithinFoundCycle(nextCycle);
         else
             try {
-                final int nextCycle0 = queue.nextCycle(this.cycle, direction);
-                if (nextCycle0 == -1)
-                    return Long.MIN_VALUE;
+                int nextCycle0 = this.cycle;
+                while (true) {
+                    nextCycle0 = queue.nextCycle(nextCycle0, direction);
+                    if (nextCycle0 == -1)
+                        return Long.MIN_VALUE;
 
-                nextIndex = nextIndexWithinFoundCycle(nextCycle0);
+                    if (cycle(nextCycle0)) {
+                        nextIndex = nextIndexWithinFoundCycle(nextCycle0);
+                        break;
+                    }
+                }
 
             } catch (ParseException e) {
                 throw new IllegalStateException(e);
@@ -1112,9 +1118,12 @@ class StoreTailer extends AbstractCloseable
         if (count <= 0)
             return false;
         final RollCycle rollCycle = queue.rollCycle();
-        moveToIndexInternal(rollCycle.toIndex(cycle, count - 1));
-        this.state = FOUND_IN_CYCLE;
-        return true;
+        if (moveToIndexInternal(rollCycle.toIndex(cycle, count - 1))) {
+            this.state = FOUND_IN_CYCLE;
+            return true;
+        } else {
+            return false;
+        }
     }
 
     void index0(final long index) {
--- a/src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
+++ b/src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
@@ -38,18 +38,24 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.OptionalLong;
+import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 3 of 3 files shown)


src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -226,6 +228,55 @@ public void tailingThroughDeletedCyclesWillRefreshThenRetry_ReadOnly() throws IO
                 .build());
     }
 
+    @Test
+    public void deletingFilesChaosTest() throws InterruptedException {
```
alamar: Does it actually do any assertions? What does it check for?

nicktindall: It's not a full fledged test yet, it just fails when an exception is thrown

nicktindall: We should be able to handle roll cycles going missing mostly gracefully, but interesting to understand the circumstances that we don't cover

alamar:
On 5.22 this test will fail every time, with `Unable to progress to the next cycle, state=END_OF_CYCLE`
If the deleter thread is removed it will pass.

5.24ea will fail around half of the runs.

Please clarify what are action points here @nicktindall

nicktindall:
> On 5.22 this test will fail every time, with `Unable to progress to the next cycle, state=END_OF_CYCLE`
> If the deleter thread is removed it will pass.
> 
> 5.24ea will fail around half of the runs.
> 
> Please clarify what are action points here @nicktindall 

Thanks for looking into that @alamar, I will do a bit more digging today, the latest being more resilient than 22 is the expected outcome, I will investigate more specific usage patterns that might have become less resilient (as reported by client).

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1216,6 +1216,10 @@ public int nextCycle(int currentCycle, @NotNull TailerDirection direction) {
                 fileFound |= currentCycleFile.exists();
 
                 if (!fileFound) {
+                    if (currentCycle < directoryListing.getMinCreatedCycle()) {
+                        Jvm.warn().on(SingleChronicleQueue.class, "Current cycle has been deleted, resuming from first available cycle");
+                        return toCycle(tree.firstEntry());
+                    }
```
nicktindall, line 1222: In the specific case that we're on a roll cycle earlier than the minimum one, and the cycle we're on is no longer there, can we skip forward to the earliest existing cycle? We probably also need to handle the case where there are all of a sudden no roll cycles. I'll beef up the testing around that.

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -361,7 +483,7 @@ public void onReleased(int cycle, File file) {
 
         @Override
         public void onAcquired(int cycle, File file) {
-            System.out.println("onAcquired called cycle: " + cycle + ", file: " + file);
+//            System.out.println("onAcquired called cycle: " + cycle + ", file: " + file);
```
JerryShea: should remove?

JerryShea (approved):
This is a really good improvement for a particular customer's use case but @alamar these are good questions

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1216,13 +1216,21 @@ public int nextCycle(int currentCycle, @NotNull TailerDirection direction) {
...
 
                 if (!fileFound) {
-                    directoryListing.refresh(true);
-                    throw new MissingStoreFileException(
-                            String.format("Expected file to exist for cycle: %d, file: %s.%nminCycle: %d, maxCycle: %d%n" +
-                                            "Available files: %s",
-                                    currentCycle, currentCycleFile,
-                                    directoryListing.getMinCreatedCycle(), directoryListing.getMaxCreatedCycle(),
-                                    Arrays.toString(path.list((d, n) -> n.endsWith(SingleChronicleQueue.SUFFIX)))));
+                    Jvm.error().on(SingleChronicleQueue.class, "The current cycle seems to have been deleted from under the queue, scanning to find the next remaining cycle");
```
alamar: May lead to data loss going unnoticed by our users...

JerryShea: Maybe we should make this behaviour configurable?

nicktindall: https://www.joelonsoftware.com/2000/04/12/choices/

alamar: That's a change of behavior and a quite significant one at that.

nicktindall: Yes,  but I think we should either decide to do it or not, I'm not a fan of adding another config option. It just adds to an already large config space meaning another configuration that isn't comprehensively tested.
And more documentation describing a fairly esoteric detail.

alamar: I would prefer to not do it, because users can work around failures but they cannot work around silent data loss caused by suppression of failures.
But that's just my opinion.

nicktindall: A couple of things about the change
- It does log an error to the logs when it goes into this recovery mode, so not completely "silent"/"suppressed"
- This recovery only occurs when you delete the current AND next roll cycle, if you delete EITHER the current or the next cycle the existing code will silently continue, so it's not like there's any protection against data loss in the existing behaviour that's being removed
- If you don't want to lose data you probably shouldn't be deleting files from under the program

src/main/java/net/openhft/chronicle/queue/TailerDirection.java
```diff
@@ -31,4 +82,32 @@ public enum TailerDirection {
...
+
+    /**
+     * Get the next entry in this direction
+     *
+     * @param cyclesTree The cycles tree
+     * @param current    The current timestamp
+     * @return The next entry in sequence after the current timestamp
+     */
+    @Nullable
+    abstract public Map.Entry<Long, File> nextEntry(NavigableMap<Long, File> cyclesTree, Long current);
```
alamar: I believe that TailerDirection should be opaque to end user and should not expose these methods that are not expected to be called by user.

alamar (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1222,42 +1221,32 @@ public int nextCycle(int currentCycle, @NotNull TailerDirection direction) {
...
-            // already checked that the file should be on-disk, so if it is null, call cycleTree again with force
             if (file == null) {
                 tree = cycleTree(true);
                 file = tree.get(key);
             }
+
+            // The current cycle is no longer on disk, log an error
             if (file == null) {
-                throw new MissingStoreFileException("missing currentCycle, file=" + currentCycleFile);
+                Jvm.error().on(SingleChronicleQueue.class, "The current cycle seems to have been deleted from under the queue, scanning to find the next remaining cycle, currentCycle=" + currentCycleFile);
```
nicktindall, line 1249: Use the navigable map to determine what the next cycle is, even if the current cycle isn't present (it may have been deleted)
Log an error when we see this happen because it's not something people should do.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -536,11 +536,17 @@ private long nextIndexWithNextAvailableCycle0(final int cycle) {
...
+                int nextCycle0 = this.cycle;
+                while (true) {
+                    nextCycle0 = queue.nextCycle(nextCycle0, direction);
+                    if (nextCycle0 == -1)
+                        return Long.MIN_VALUE;
+
+                    if (cycle(nextCycle0)) {
+                        nextIndex = nextIndexWithinFoundCycle(nextCycle0);
+                        break;
+                    }
```
nicktindall, line 548: If the next cycle found (according to the caches) can't be loaded, keep traversing forwards until we find one that can (or we reach the end)

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -1112,9 +1118,12 @@ private boolean tryWindBack(final int cycle) {
...
         final RollCycle rollCycle = queue.rollCycle();
-        moveToIndexInternal(rollCycle.toIndex(cycle, count - 1));
-        this.state = FOUND_IN_CYCLE;
-        return true;
+        if (moveToIndexInternal(rollCycle.toIndex(cycle, count - 1))) {
+            this.state = FOUND_IN_CYCLE;
+            return true;
+        } else {
+            return false;
+        }
```
nicktindall, line 1126: ONLY set the state to `FOUND_IN_CYCLE` after we've successfully loaded the cycle, we can't assume that because we were able to find the last index a moment ago, that the cycle is still there and loadable.

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -192,38 +198,279 @@ public void firstAndLastIndexAreRefreshedAfterForceRefreshInterval() throws IOEx
...
+
+            backwardTailerThread.start();
+            forwardTailerThread.start();
+            deleterThread.start();
+            deleterThread.join();
+            running.set(false);
+            forwardTailerThread.join();
+            backwardTailerThread.join();
+        }
+    }
```
nicktindall, line 252: This test progressively deletes cycles in batches from the beginning of the queue. It probably mirrors the way sensible people archive old roll cycles.

While that's occurring, we have two threads continually scanning over the queue in both directions and validating that they saw all the remaining roll cycles. (this is a bit hard to check so instead we ensure they saw all the roll cycles not marked as deleted at the moment they complete their traversal of the queue).

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -192,38 +198,279 @@ public void firstAndLastIndexAreRefreshedAfterForceRefreshInterval() throws IOEx
...
+
+            backwardTailerThread.start();
+            forwardTailerThread.start();
+            deleteRandomCyclesThread.start();
+            deleteRandomCyclesThread.join();
+            running.set(false);
+            backwardTailerThread.join();
+            forwardTailerThread.join();
+        }
+    }
```
nicktindall, line 365: This test randomly deletes cycles at any point in the queue. It probably doesn't mirror any kind of sensible archiving strategy, it was more to add confidence that Tailers are still able to traverse under chaotic conditions.

Again, while that's occurring, we have two threads continually scanning over the queue in both directions and validation that they saw all the remaining roll cycles. (this is a bit hard to check so we ensure they saw all the roll cycles not deleted when they reached the start/end of the queue).

peter-lawrey (approved):

PR #1374, opened 2023-06-20
Deprecate acquire appender

nicktindall:
Also made a start on removing its use from tests, but that's a bigger job. Just wanted to get some coverage of it being used. I would remove all the low-hanging-fruit uses of `acquireAppender` from tests before merging if this gets approved.

Also will follow with a PR to update demos once it's been released.

I also changed it so that Appenders that are `acquire`'d will throw `UnsupportedOperationException` when `singleThreadCheck(Disabled|Reset)` is called. This needs [a corresponding change](https://github.com/OpenHFT/Chronicle-Core/pull/532) in core to ignore that exception when we do `net.openhft.chronicle.core.io.AbstractCloseable#waitForCloseablesToClose`.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -505,6 +505,17 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
         return res;
     }
 
+    @NotNull
+    @Override
+    public ExcerptAppender createAppender() {
+        throwExceptionIfClosed();
+
+        if (readOnly)
+            throw new IllegalStateException("Can't append to a read-only chronicle");
+
+        return createNewAppenderOnceConditionIsMet();
+    }
+
     /**
      * @return the {@link QueueLock} This lock is held while the queue replication cluster is back-filling.
      * By Back-filling we mean that, as part of the fail-over process a sink, may actually have more data than a source,
--- a/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
@@ -194,9 +194,29 @@ public interface ChronicleQueue extends Closeable {
      * this method every time an appender is needed.
      *
      * @return Returns a ExcerptAppender for this ChronicleQueue that is local to the current Thread
+     * @deprecated It is recommended to use {@link #createAppender()} instead
      */
+    @Deprecated(/* To be removed in x.TBD */)
     @NotNull
     ExcerptAppender acquireAppender();
+
+    /**
+     * Creates and returns a new ExcerptAppender for this ChronicleQueue
+     * <p>
+     * An Appender can be used to store new excerpts sequentially to the queue.
+     * <p>
+     * <b>
+     * An Appender is <em>NOT thread-safe</em>, concurrent use of an Appender by multiple threads is unsafe
+     * and will inevitably lead to errors and unspecified behaviour.
+     * </b>
+     * <p>
+     * This method creates a new Appender on each call, it is up to the caller to manage the lifecycle of the
+     * returned Appender
+     *
+     * @return Returns a new ExcerptAppender for this ChronicleQueue
+     */
+    @NotNull
+    ExcerptAppender createAppender();
 
     /**
      * Returns the lowest valid index available for this ChronicleQueue, or {@link Long#MAX_VALUE}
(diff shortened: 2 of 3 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -135,7 +135,12 @@
     private int cycleFileRenamed = -1;
     @NotNull
     private Condition createAppenderCondition = NoOpCondition.INSTANCE;
-    protected final ThreadLocal<ExcerptAppender> strongExcerptAppenderThreadLocal = CleaningThreadLocal.withCloseQuietly(this::createNewAppenderOnceConditionIsMet);
+    protected final ThreadLocal<ExcerptAppender> strongExcerptAppenderThreadLocal = CleaningThreadLocal.withCloseQuietly(() -> {
+        final StoreAppender newAppenderOnceConditionIsMet = (StoreAppender) this.createNewAppenderOnceConditionIsMet();
+        // Thread-local Appenders should prevent single-threaded checks being disabled/reset
+        newAppenderOnceConditionIsMet.bindToCreatingThread();
```
nicktindall, line 141: This needs a corresponding change in core to ignore UnsupportedOperationException when `net.openhft.chronicle.core.io.AbstractCloseable#waitForCloseablesToClose` calls `singleThreadCheckDisabled` on the appender.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -505,6 +510,17 @@ public ExcerptAppender acquireAppender() {
...
 
+    @NotNull
+    @Override
+    public ExcerptAppender createAppender() {
+        throwExceptionIfClosed();
+
+        if (readOnly)
+            throw new IllegalStateException("Can't append to a read-only chronicle");
+
+        return createNewAppenderOnceConditionIsMet();
```
nicktindall, line 516: `createAppender` will also wait for the `createAppenderCondition` to be met before returning (a-la `acquireAppender`)

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -138,6 +139,24 @@ class StoreAppender extends AbstractCloseable
...
+            throw new UnsupportedOperationException("This appender cannot be used by any thread but the creating thread");
+        super.singleThreadedCheckReset();
+    }
+
+    @Override
+    public void singleThreadedCheckDisabled(boolean singleThreadedCheckDisabled) {
+        if (boundToCreatingThread)
+            throw new UnsupportedOperationException("This appender cannot be used by any thread but the creating thread");
+        super.singleThreadedCheckDisabled(singleThreadedCheckDisabled);
+    }
```
nicktindall, line 158: This allows thread-local Appenders to prevent bypassing of the single thread protections.

src/test/java/net/openhft/chronicle/queue/AcquireReleaseTest.java
```diff
@@ -69,7 +69,7 @@ public void onReleased(int cycle, File file) {
                 .build()) {
 
             int iter = 4;
-            try (ExcerptAppender excerptAppender = queue.acquireAppender()) {
+            try (ExcerptAppender excerptAppender = queue.createAppender()) {
```
nicktindall, line 72: All the instances that look like this are a find and replace, there are some cases where it's less simple

src/test/java/net/openhft/chronicle/queue/ChronicleQueueTwoThreadsTest.java
```diff
@@ -54,6 +53,7 @@ void doTest(boolean buffered) throws InterruptedException {
             Bytes<?> bytes = NativeBytes.nativeBytes(BYTES_LENGTH).unchecked(true);
             try (ChronicleQueue rqueue = SingleChronicleQueueBuilder
                     .fieldlessBinary(name)
+                    .rollCycle(SMALL_DAILY)
```
nicktindall, line 56: This was inconsistent with the block below

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
nicktindall: Actually had to disable this because `PublishHandler` (some part of channels) attempts to disable thread safety checks on Appenders for some reason.

I would like to put it in though.

JerryShea (approved):

PR #1401, opened 2023-08-01
Refactor TableStoreWriteLock.lock to reduce length fixes #1400

### LINKED ISSUE #1400: TableStoreWrite.lock reduce method length
Simplify this method and reduce its length to bring its bytecode length under 325 bytes (-XX:FreqInlineSize=325) to give it a better chance of being inlined during JIT compilation.

# Before

Before this change the `lock` method had a bytecode length greater than 325 bytes. A benchmark was setup to run an appender in a tight loop ensuring sufficient time for compilation. The compilation log was captured including inlining information. The only methods that were inlined from within the `TableStoreWriteLock` class are shown below - note that `lock` was *not* inlined:

before-optimisation

# After

After the refactor of the `lock` method and its new associated sub-methods the same benchmark was re-run and we can observe `lock` being appropriately inlined:

compilation-output-inlining

The output of jclasslib showing the length of the bytecode is shown below:

updated-method-length

tgd:
Please see context / write up here - https://github.com/OpenHFT/Chronicle-Queue/issues/1400

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
@@ -69,38 +69,55 @@ public class TableStoreWriteLock extends AbstractTSQueueLock implements WriteLoc
         try {
             currentLockValue = lock.getVolatileValue();
             while (!lock.compareAndSwapValue(UNLOCKED, PID)) {
-                if (Thread.currentThread().isInterrupted())
-                    throw new InterruptedRuntimeException("Interrupted for the lock file:" + path);
-                tlPauser.pause(timeout, TimeUnit.MILLISECONDS);
-                currentLockValue = lock.getVolatileValue();
+                currentLockValue = lockGetCurrentLockValue(tlPauser);
             }
 
-            //noinspection ConstantConditions,AssertWithSideEffects
-            assert (lockedByThread = Thread.currentThread()) != null
-                    && (lockedHere = new StackTrace()) != null;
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -69,40 +69,57 @@ public void lock() {
...
+    private long lockGetCurrentLockValue(TimingPauser tlPauser) throws TimeoutException {
+        if (Thread.currentThread().isInterrupted())
+            throw new InterruptedRuntimeException("Interrupted for the lock file:" + path);
+        tlPauser.pause(timeout, TimeUnit.MILLISECONDS);
+        return lock.getVolatileValue();
+    }
+
+    private void lockAssertPostConditions() {
+        //noinspection ConstantConditions,AssertWithSideEffects
+        assert (lockedByThread = Thread.currentThread()) != null
```
JerryShea, line 94: consider using `assert AssertUtil.SKIP_ASSERTIONS || ` which is a further watered-down flavour of assertion that we use

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
tgd: Thanks for the tip Jerry - looks like this assertion lib is not on the classpath of Queue at the moment - let me add it in a separate PR.

PR #1413, opened 2023-08-08
LastAcknowledgedIndexReplicatedStrategy

JerryShea:


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -292,6 +294,14 @@ public void lastAcknowledgedIndexReplicated(long newValue) {
             lastAcknowledgedIndexReplicated.setMaxValue(newValue);
     }
 
+    @Override
+    public void lastAcknowledgedIndexReplicated(long acknowledgeIndex, int remoteHostId) {
+        if (lastAcknowledgedIndexReplicated != null) {
+            long newValue = lastAcknowledgedIndexReplicatedStrategy.onAckReceived(acknowledgeIndex, remoteHostId);
```
benbonavia: this lastAcknowledgedIndexReplicationStrategy needs its own null checking -> should it be within this if block of along side it?

src/main/java/net/openhft/chronicle/queue/impl/single/LastAcknowledgedIndexReplicatedStrategy.java
```diff
@@ -0,0 +1,5 @@
+package net.openhft.chronicle.queue.impl.single;
+
+public interface LastAcknowledgedIndexReplicatedStrategy {
```
tgd, line 9: Please add javadoc

tgd (changes requested):

src/main/java/net/openhft/chronicle/queue/impl/single/LastAcknowledgedIndexReplicatedStrategy.java
tgd: What is the acceptable range of values that `onAckReceived` can return?

JerryShea:
Approved by me

tgd (approved):

benbonavia (approved):

PR #1414, opened 2023-08-10
Refactor StoreTailer.readingDocument

tgd:
Refactoring made during benchmarking session focusing on tailer throughput. The following differences were observed between this branch and develop during benchmarking:

Case | L1 loads | L1 load misses | Total branch misses
-- | -- | -- | --
develop | 141000 | 17000 | 25.7%
feature branch | 141000 | 15000 | 22.64%

This feature branch does the following:

* Adds some javadoc
* Cleans up some methods and encapsulates some logic
* Moves some code off the hot path

https://github.com/OpenHFT/Chronicle-Queue/issues/1415

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -18,6 +18,7 @@
 
 package net.openhft.chronicle.queue.impl.single;
 
+import net.openhft.chronicle.assertions.AssertUtil;
 import net.openhft.chronicle.bytes.Bytes;
 import net.openhft.chronicle.bytes.MappedBytes;
 import net.openhft.chronicle.bytes.MappedBytesStore;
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -228,16 +236,14 @@ DocumentContext readingDocument0(final boolean includeMetaData) {
...
 
-            Wire wire = context.wire();
-            if (wire != null && context.present(next)) {
-                Bytes<?> bytes = wire.bytes();
-                context.setStart(bytes.readPosition() - 4);
-                readingDocumentFound = true;
-                this.lastReadIndex = this.index();
-                return context;
+            // An entry has been found, prepare the context and return it.
+            if (next && (context.wire() != null)) {
```
nicktindall, line 240: I notice we went from `if (wire != null && context.present(next))` to `if (next && context.wire() != null)` I can see that `context.present(...)` just returns the value passed in but it also has the side-effect :vomiting_face: of setting the `present` field in the context.

Is that change significant?

nicktindall: Oh I see it happens in init now. Much nicer.

nicktindall (approved):

Issue #1416, opened 2023-08-11
AbstractTSQueueLock improve warn logging during force unlock


tgd:
We have received reports that forced queue unlocks are not visible in production because we only log notification of this at debug level. Raise this to `warn` so that it is more visible to users in production.

hft-team-city:
Released in [Chronicle-Queue-5.24ea26](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.24ea26), [BOM-2.24ea78](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24ea78)

PR #1417, opened 2023-08-11
AbstractTSQueueLock log warn on force unlock

tgd:
See https://github.com/OpenHFT/Chronicle-Queue/issues/1416 for context

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/QueueTestCommon.java
+++ b/src/test/java/net/openhft/chronicle/queue/QueueTestCommon.java
@@ -131,7 +131,7 @@ public class QueueTestCommon {
         exceptionTracker = JvmExceptionTracker.create(false);
         if (OS.isWindows())
             ignoreException("Read-only mode is not supported on Windows® platforms, defaulting to read/write");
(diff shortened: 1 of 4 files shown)


rogersimmons (approved):

src/test/java/net/openhft/chronicle/queue/QueueTestCommon.java
```diff
@@ -131,7 +131,7 @@ public void recordExceptions() {
         exceptionTracker = JvmExceptionTracker.create(false);
         if (OS.isWindows())
             ignoreException("Read-only mode is not supported on Windows® platforms, defaulting to read/write");
-        for (String msg : "Shrinking ,Allocation of , ms to add mapping for ,jar to the classpath, ms to pollDiskSpace for , us to linearScan by position from ,File released ,Overriding roll length from existing metadata, was 3600000, overriding to 86400000   ".split(",")) {
+        for (String msg : "Forced unlock, Shrinking ,Allocation of , ms to add mapping for ,jar to the classpath, ms to pollDiskSpace for , us to linearScan by position from ,File released ,Overriding roll length from existing metadata, was 3600000, overriding to 86400000   ".split(",")) {
```
JerryShea: I don't think this should be a blanket addition. It should not happen too often in the tests surely?

tgd: Good call Jerry - that's changed

PR #1418, opened 2023-08-13
review TODO comments, deprecate

JerryShea:


nicktindall (approved):

src/test/java/net/openhft/chronicle/queue/DirectoryUtils.java
```diff
@@ -67,7 +68,7 @@ enum DeleteStatic {
 
         {
             // TODO: should not need to do this now
```
tgd, line 70: Do we still need this todo?

tgd (commented):
LGTM

src/test/java/net/openhft/chronicle/queue/DirectoryUtils.java
JerryShea: Not sure ;)

PR #1422, opened 2023-08-25
Remove a check duplicated by the QueueTestCommon. 

peter-lawrey:
This check could cause the test to time out for multiple queues.

The QueueTestCommon already does this check for all resources at the end of the test, but checking each queue serially as well could cause a timeout. In at least one case, a timeout occurred in this check.


RobAustin (commented):
We are now missing the output of "      .on(getClass(), "Took " + i + " ms to release " + retained);" is this intended

RobAustin (changes requested):
its failing to pass tests on ARM

RobAustin (approved):

PR #1427, opened 2023-08-25
Make inner timeouts shorter than the overall timeout

peter-lawrey:


src/test/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLockTest.java
```diff
@@ -242,7 +242,7 @@ public void forceUnlockIfProcessIsDeadWillFailWhenLockingProcessIsAlive() throws
             assertTrue(lock.locked());
         }
         lockingProcess.destroy();
-        lockingProcess.waitFor(5_000, TimeUnit.SECONDS);
+        lockingProcess.waitFor(3_000, TimeUnit.SECONDS);
```
RobAustin, line 245: This change could cause tests to break on slower platforms

src/test/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLockTest.java
```diff
@@ -252,7 +252,7 @@ public void forceUnlockIfProcessIsDeadWillSucceedWhenLockingProcessIsDead() thro
         try (TableStoreWriteLock lock = createTestLock()) {
             waitForLockToBecomeLocked(lock);
             lockingProcess.destroy();
-            lockingProcess.waitFor(5_000, TimeUnit.SECONDS);
+            lockingProcess.waitFor(3_000, TimeUnit.SECONDS);
```
RobAustin, line 255: This change could cause tests to break on slower platforms

src/test/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLockTest.java
```diff
@@ -269,7 +269,7 @@ public void forceUnlockIfProcessIsDeadWillSucceedWhenLockIsNotLocked() {
     private void waitForLockToBecomeLocked(TableStoreWriteLock lock) throws TimeoutException {
         Pauser p = Pauser.balanced();
         while (!lock.locked()) {
-            p.pause(5_000, TimeUnit.SECONDS);
+            p.pause(3_000, TimeUnit.SECONDS);
```
RobAustin, line 272: This change could cause tests to break on slower platforms.

RobAustin (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLockTest.java
RobAustin: for example, it is not passing on Linux ARM currently

Issue #1431, opened 2023-08-25
Method change in TableStoreWriteLock causes spurious test failure


alamar:
```
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] [main] INFO software.chronicle.enterprise.queue.replication.stress.UngracefulShutdownTest - Exception detected in process: producer
...
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 2023-08-25T23:37:10.863+0100 [main] WARN net.openhft.chronicle.queue.impl.single.TableStoreWriteLock - Forced unlock for the lock file:replica/host1/source/queues/queue/metadata.cq4t, lockKey: chronicle.write.lock, unlocked: true
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] net.openhft.chronicle.core.StackTrace: Forced unlock on main
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.table.AbstractTSQueueLock.forceUnlock(AbstractTSQueueLock.java:83)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.single.TableStoreWriteLock.lockHandleTimeoutException(TableStoreWriteLock.java:112)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.single.TableStoreWriteLock.lock(TableStoreWriteLock.java:81)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.single.StoreAppender.prepareAndReturnWriteContext(StoreAppender.java:429)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.single.StoreAppender.writingDocument(StoreAppender.java:413)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at net.openhft.chronicle.queue.impl.single.StoreAppender.writingDocument(StoreAppender.java:401)
[22:37:11]W:		 [software.chronicle:chronicle-queue-enterprise] 	at software.chronicle.enterprise.queue.replication.stress.common.Producer.writeToQueue(Producer.java:84)
[...]

PR #1432, opened 2023-08-29
TableStoreWriteLock rename exception handling method

tgd:
Resolve spurious test failure - see #1431 for more details.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
@@ -77,7 +77,7 @@ public class TableStoreWriteLock extends AbstractTSQueueLock implements WriteLoc
 
             // success
         } catch (TimeoutException e) {
-            lockHandleTimeoutException(currentLockValue);
+            lockHandleTimeout(currentLockValue);
         } finally {
             tlPauser.reset();
         }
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -96,7 +96,7 @@ private void lockAssertPostConditions() {
                 ((lockedByThread = Thread.currentThread()) != null && (lockedHere = new StackTrace()) != null);
     }
 
-    private void lockHandleTimeoutException(long currentLockValue) {
+    private void lockHandleTimeout(long currentLockValue) {
```
alamar: Perhaps add some verb to method name?

alamar (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
tgd: How about `handleLockTimeout`?

alamar: Looks OK if it actually does that!

tgd: It does. I tweaked the naming a little bit more - I also removed the `Exception` substring from another method!

alamar (approved):

PR #1435, opened 2023-08-31
Test appears to have not finishedNormally so the reference check obscures the real reason.

peter-lawrey:


src/test/java/net/openhft/chronicle/queue/impl/single/SingleCQFormatTest.java
```diff
@@ -231,6 +232,7 @@ public void testCompleteHeader() throws FileNotFoundException {
         } catch (Exception e) {
             e.printStackTrace();
```
RobAustin, line 233: should this be e.printStackTrace() ? or JVM logging

RobAustin (approved):

PR #1438, opened 2023-09-07
Multiple empty cycle files with no EOF markers and pretoucher may cause tailer stall 

RobAustin:
relates to https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/563

Introduction

When a queue contains multiple empty cycle files that (erroneously) contain no EOF markers then any tailer beginning to read from the beginning of the queue will stall in the first empty cycle without the EOF marker.

Two key things have to align in order to reproduce this issue:

The pretoucher must be in use - this is what will ensure that empty cycle files are created
There must be at least 2 cycles that do not contain any data
Impacted branches

develop, ea, release.*
Steps to reproduce

Sample code to reproduce this can be found here: https://github.com/OpenHFT/Chronicle-Queue/pull/1437/files#diff-86a449a50c5152bc5ef5198d1659d8720c2f5ffd667991ced965b173345edb00

Create a queue using a roll cycle that will create many cycles for testing purposes (e.g. TEST_SECONDLY)
Instantiate a pretoucher that will pretouch this queue in the background and will also precreate cycle files that do not already exist
Create an appender that will append to this queue
Append some initial data
Wait for some period of time until multiple rolls have occurred
Ensure that for at least 2 rolls you do not append any data to the queue
Append another entry in the queue - you will never be able to reach this entry with the tailer
Next, create a tailer and tail through the queue from the beginning. The tailer should get stuck on one of the cycle files.
Background

A large amount of work has gone into the normaliseEOFs feature to ensure that all cycles contain a valid EOF marker
In this particular scenario we hit a specific code path where the EOF marker is not stamped in successive cycle files
Workarounds

It is possible to move the tailer on by calling moveToIndex to move it past the cycle on which it is stuck.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -779,6 +779,9 @@ class StoreAppender extends AbstractCloseable
                 for (int eofCyle = cycle - 1; eofCyle >= queue.firstCycle(); eofCyle--) {
 
                     try (SingleChronicleQueueStore st = queue.storeForCycle(eofCyle, queue.epoch(), false, null)) {
+
+                        if (st == null)
+                            continue;
                         wire = queue.wireType().apply(st.bytes());
                         wire.usePadding(st.dataVersion() > 0);
                         if (!st.writeEOF(wire, timeoutMS())) {
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -772,7 +772,25 @@ class StoreAppender extends AbstractCloseable
 
         if (!suppressEOF) {
             assert queue.writeLock().locked();
-            store.writeEOF(wire, timeoutMS());
+
+            if (store.writeEOF(wire, timeoutMS())) {
+                Wire w = this.wire;
+
+                for (int eofCyle = cycle - 1; eofCyle >= queue.firstCycle(); eofCyle--) {
+
+                    try (SingleChronicleQueueStore st = queue.storeForCycle(eofCyle, queue.epoch(), false, null)) {
+                        wire = queue.wireType().apply(st.bytes());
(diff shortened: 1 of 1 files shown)


RobAustin:
@peter-lawrey suggested writing in the EOF in a background thread rather than using the appender; 
1. This would add a fair amount of complexity
2. The appender.pretouch() could be called in the same thread as the appender
3. In order to do this safely, we would have to lock the appenders anyway, stalling the appender

JerryShea (changes requested):
Consider to instead call normaliseEOF

JerryShea (changes requested):
Let's please discuss some more before applying this

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -772,7 +772,28 @@ private void rollCycleTo(final int cycle, boolean suppressEOF) {
 
         if (!suppressEOF) {
             assert queue.writeLock().locked();
-            store.writeEOF(wire, timeoutMS());
+
+            if (store.writeEOF(wire, timeoutMS())) {
+                Wire w = this.wire;
+
+                for (int eofCyle = cycle - 1; eofCyle >= queue.firstCycle(); eofCyle--) {
```
alamar, line 779: I believe we already have it somewhere (normalizeEOFs) and it would be great to figure out why it did not work.

RobAustin: Sure - you can run the test case and see. The test case is in ChronicleQueueEnterpise. see https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/563

nicktindall: I'd prefer we [fixed the root cause](https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/issues/423) rather than adding this complexity, also this will make rolling very expensive, probably negating any gains had by enabling `earlyAcquireNextCycle`.

PR #1442, opened 2023-09-26
disable Pretroucher canWrite

JerryShea:


tgd:
Hi Jerry

Here are some unit tests from the old PR for this in case you find them
useful -
https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/568/files

Thanks

Tom

PR #1443, opened 2023-09-26
WIP: use weak references for Queue close listeners

JerryShea:
Experiment with weak references in the Queue close-listener registry so that registering an appender or tailer does not itself retain that object. The diff also adds explicit resource closure in several tests and introduces a close-listener removal overload.

The current source has the appender/tailer calls to remove themselves from the registry commented out. The earlier description below records a two-part proposal; automatic removal on close is not delivered by this head.

WIP - do not merge. The discussion records failing or flaky leak checks after earlier revisions. Correct and test the weak-reference pruning predicate, including live and collected referents, then verify queue closure, explicit closure and leak detection together. Do not suppress discarded-resource warnings to make those checks pass.

Validation: the diff and recorded review/check history were inspected at `53d0e0e6d26b`. No build, test or benchmark was rerun for this metadata edit. No CI result was returned for this head.

---

## Existing author and review notes

WIP - **do not merge**

Fixes two problems:
1. makes sure that tailers and appenders remove themselves from SingleChronicleQueue.closers
2. makes closers a WeakReference so that the SCQ is no longer the only thing holding a reference to tailers & appenders after they would otherwise be candidates for GC

Second point above we should definitely discuss.

If we can reach consensus on this then after this change has been merged, it will be possible to set `disable.discard.warning=false` to register a finalizer for both appenders and tailers and ensure that if the finalizer is called, that they will warn and close themselves, thus releasing any resources. Backporting this to 23 will not be trivial as discard warnings/finalizers were different back then.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
@@ -47,6 +47,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.*;
+import java.lang.ref.WeakReference;
 import java.security.SecureRandom;
 import java.text.ParseException;
 import java.time.ZoneId;
@@ -110,7 +111,7 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
     private final TimeProvider time;
     @NotNull
     private final BiFunction<RollingChronicleQueue, Wire, SingleChronicleQueueStore> storeFactory;
-    private final Set<Closeable> closers = Collections.newSetFromMap(new IdentityHashMap<>());
+    private final Set<WeakReference<Closeable>> closers = Collections.newSetFromMap(new IdentityHashMap<>());
     private final boolean readOnly;
     @NotNull
     private final CycleCalculator cycleCalculator;
@@ -716,8 +717,11 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
     public <T> void addCloseListener(Closeable key) {
         synchronized (closers) {
             if (!closers.isEmpty())
-                closers.removeIf(Closeable::isClosed);
-            closers.add(key);
+                closers.removeIf(wrc -> {
+                    final Closeable closeable = wrc.get();
+                    return closeable != null || closeable.isClosed();
+                });
+            closers.add(new WeakReference<>(key));
         }
     }
 
@@ -725,7 +729,7 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
     @Override
     protected void performClose() {
         synchronized (closers) {
-            metaStoreMap.values().forEach(Closeable::closeQuietly);
+            Closeable.closeQuietly(metaStoreMap.values());
             metaStoreMap.clear();
             closers.forEach(Closeable::closeQuietly);
             closers.clear();
@@ -746,7 +750,7 @@ SingleChronicleQueue extends AbstractCloseable implements RollingChronicleQueue 
 
         // close it if we created it.
         if (eventLoop instanceof OnDemandEventLoop)
-            eventLoop.close();
+            Closeable.closeQuietly(eventLoop);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


JerryShea:
I reverted the first change (to call removeCloseListener from appender/tailer close methods) as a bunch of tests failed - will come back to.

The second change gives us some test flakiness with tests occasionally failed with `exceptions were detected: Discarded without closing...` and the tests that have shown this have tailers which are not explicitly closed and could go out of scope (and thus be GC'd) before the parent queue is closed. I have fixed a few of these tests to close tailers explicitly but there are a lot of them. We can 
a) fix all these tests - it would be the ideal situation for our tests to demonstrate proper lifecycle management OR
b) `ignoreException` the "Discarded without closing" warning and create an issue to fix the tests

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -209,6 +209,7 @@ public void writeBytes(@NotNull final WriteBytesMarshallable marshallable) {
 
     @Override
     protected void performClose() {
+//        queue.removeCloseListener(this);
```
JerryShea, line 228: see comments - will come back to

nicktindall:
> I reverted the first change (to call removeCloseListener from appender/tailer close methods) as a bunch of tests failed - will come back to.
> 
> The second change gives us some test flakiness with tests occasionally failed with `exceptions were detected: Discarded without closing...` and the tests that have shown this have tailers which are not explicitly closed and could go out of scope (and thus be GC'd) before the parent queue is closed. I have fixed a few of these tests to close tailers explicitly but there are a lot of them. We can a) fix all these tests - it would be the ideal situation for our tests to demonstrate proper lifecycle management OR b) `ignoreException` the "Discarded without closing" warning and create an issue to fix the tests

I vote definitely not b - unless it's done locally on the problematic tests, but if you're doing that you might as well fix them. "Creating an issue" no way guarantees it will ever get done, in the meantime, we are muting the thing we use to detect resource leaks so we will run the risk of introducing resource leaks. If it's not tested it's broken.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -732,16 +733,19 @@ public NavigableSet<Long> listCyclesBetween(int lowerCycle, int upperCycle) thro
     public <T> void addCloseListener(Closeable key) {
         synchronized (closers) {
             if (!closers.isEmpty())
-                closers.removeIf(Closeable::isClosed);
-            closers.add(key);
+                closers.removeIf(wrc -> {
+                    final Closeable closeable = wrc.get();
+                    return (closeable != null) ? closeable.isClosed() : true;
```
nicktindall: We don't need to keep empty weak references around do we? 
`return closeable == null || closeable.isClosed()`

nicktindall: Sorry that's what it does, but perhaps refactor to less byzantine logic

src/test/java/net/openhft/chronicle/queue/QueueTestCommon.java
```diff
@@ -133,6 +133,10 @@ public void recordExceptions() {
         for (String msg : "Shrinking ,Allocation of , ms to add mapping for ,jar to the classpath, ms to pollDiskSpace for , us to linearScan by position from ,File released ,Overriding roll length from existing metadata, was 3600000, overriding to 86400000   ".split(",")) {
             ignoreException(msg);
         }
+        // See https://github.com/OpenHFT/Chronicle-Queue/issues/1455. As many unit tests do not use try/finally
+        // to manage tailer or appender scope properly. we are ignoring this exception for now.
+        // TODO: remove the below line and run the tests many times to flush out the problematic tests (or else inspect the codebase)
+        ignoreException("Discarded without closing");
```
nicktindall: big one of these :-1:

nicktindall (changes requested):
Don't silence leak detection

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -104,7 +105,7 @@
     private final TimeProvider time;
     @NotNull
     private final BiFunction<RollingChronicleQueue, Wire, SingleChronicleQueueStore> storeFactory;
-    private final Set<Closeable> closers = Collections.newSetFromMap(new IdentityHashMap<>());
+    private final Set<WeakReference<Closeable>> closers = Collections.newSetFromMap(new IdentityHashMap<>());
```
peter-lawrey, line 114: Need to ensure these resources are closed if discarded, otherwise, the resources they reference won't be cleaned up except by further GCs, if they are closed on being discarded.

PR #1447, opened 2023-10-03
Change default pretoucher behaviour - back port to x.23

tgd:
Back port to x.23 release.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
@@ -156,7 +156,7 @@ public class SingleChronicleQueueTest extends ChronicleQueueTestBase {
 
     @Test
     public void createAppenderWillThrowWhenQueueIsReadOnly() {
-        assumeTrue(OS.isWindows()); // Read only is not supported on windows
+        assumeFalse(OS.isWindows()); // Read only is not supported on windows
         final File queueDir = getTmpDir();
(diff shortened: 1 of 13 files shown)


rogersimmons (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
```diff
@@ -156,6 +156,7 @@ public void createAppenderWillReturnANewAppenderEachTime() {
 
     @Test
     public void createAppenderWillThrowWhenQueueIsReadOnly() {
+        assumeFalse(OS.isWindows()); // Read only is not supported on windows
```
JerryShea, line 159: the convention generally used is
```
assumeFalse("Read-only mode is not supported on Windows", OS.isWindows());
```

JerryShea (approved):
approved with 1 minor comment

nicktindall (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueTest.java
JerryShea: @tgd I bumped CQ in 2.23 bom too so I would be very happy for you to kick off a 23 release once merged and built

JerryShea: @tgd seeking your forgiveness rather than permission to merge this, as we would like to cut a 23 release today

PR #1456, opened 2023-10-17
deprecate unused classes - they have been copied to Services together with IKVS

JerryShea:
See [build-all](https://teamcity.chronicle.software/project/Chronicle_BuildAll?branch=feature%2Fremove-unused&buildTypeTab=overview&mode=builds)


src/main/java/net/openhft/chronicle/queue/internal/domestic/backed/map/QueueBackedMaps.java
```diff
@@ -10,6 +10,7 @@
 /**
  * Used by services to handle multiple QueueBackedMaps
  */
+@Deprecated(/* to be removed in x.25 */)
```
JerryShea, line 13: These are no longer used in the codebase, and were only used by legacy QBM which was deprecated before and has now been removed, so appropriate to remove in x.25

peter-lawrey (approved):

PR #1460, opened 2023-10-20
Named tailer index version + set up preconditions for named tailer creation

tgd:


src/test/java/net/openhft/chronicle/queue/impl/single/RollingCycleTest.java
```diff
@@ -42,6 +43,7 @@
 import static net.openhft.chronicle.queue.rollcycles.TestRollCycles.TEST_DAILY;
 import static org.junit.Assert.*;
 
+@Ignore("Temporarily ignored as binary contents of metadata store has changed, this test needs to be refactored")
```
tgd: This test needs fixing/refactoring prior to merge

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -547,23 +547,43 @@ public WriteLock appendLock() {
...
 
+    private void verifyTailerPreconditions(String id) {
+        // Preconditions for all tailer types
+        throwExceptionIfClosed();
+
+        // Named tailer preconditions
+        if (id == null) return;
+        if (appendLock.locked()) {
+            throw new NamedTailerNotAvailableException(id, NamedTailerNotAvailableException.Reason.NOT_AVAILABLE_ON_SINK);
+        }
```
nicktindall, line 570: Should this be more specific and check that the tailer is not a replicated named tailer. I don't think we want to remove the ability to use non-replicated named tailers on sinks.

I wonder if we could designate that replicated named tailers start with a prefix or something? `replicated:somename` instead of having to specify them in a config somewhere.

nicktindall (approved):
Looking good, I was just concerned about retaining the ability to use non-replicated named tailers on sinks, as I suspect there will be many people doing that (as they're a consumer thing)

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
tgd: Hi @nicktindall good spot. You are right - this needs to change. I will follow the `replicated:somename` approach and update the PR

tgd: This has now been updated

rogersimmons (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -547,23 +548,43 @@ public WriteLock appendLock() {
...
     @Override
     @NotNull
     public LongValue indexForId(@NotNull String id) {
         return this.metaStore.doWithExclusiveLock((ts) -> ts.acquireValueFor("index." + id, 0L));
     }
 
+    @NotNull
+    public LongValue indexVersionForId(@NotNull String id) {
+        return this.metaStore.doWithExclusiveLock((ts) -> ts.acquireValueFor("index." + id + ".version", -1L));
+    }
```
nicktindall, line 582: We don't really need to have a version for non-replicated named tailers. I wonder if it's worth having an abstraction e.g. IndexUpdater and it bumps the index and the version for a replicated tailer, but only writes the index for a non-replicated one. It might be easier to just have them both write the version? up to you.

src/test/java/net/openhft/chronicle/queue/impl/single/RollingCycleTest.java
```diff
@@ -295,6 +304,7 @@ public void testRollCycle() {
             // System.out.println("Wrote " + numWritten + " Read " + numRead);
 
             String dump = queue.dump();
+            System.out.println(dump);
```
nicktindall: Left in by accident?

nicktindall (approved):
Approved but consider whether we need version for non-replicated Tailer and whether there's any advantage to omitting it.

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
tgd: Hi @nicktindall thanks for this - I think you're right and that this would be an improvement. I have added the `IndexUpdater` and `IndexUpdaterFactory` classes that implement this. This does seem cleaner and minimises the impact to existing named tailers. Moreover we avoid the cost of the version increment in non replicated tailers.

tgd: In the future I would like to come back and resolve all of the null checking in StoreTailer - for example a null index was historically passed into StoreTailer to indicate an unnamed tailer. The new additions still conform to this design of passing null around. I would like to come back and refactor this after we have released this.

PR #1462, opened 2023-10-30
Remove deprecated QueueLock Fixes #1461

### LINKED ISSUE #1461: Remove QueueLock
The QueueLock was used to prevent writes to the queue while a back-fill was in progress. A better approach was implemented via the replication heartbeat, this indicates to potential writers that the replicator has completed all back-fills and it is safe to write. Writers can choose whether to only write when safe, or write at other times when they know it to be safe.

nicktindall:
I believe this is safe now, we have removed usage of it in replication (see https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/508)


JerryShea (approved):
nice clean up

yevgenp (approved):

rogersimmons (approved):

tgd (approved):

PR #1467, opened 2023-11-06
Update binary compatibility reference.

yevgenp:


tgd (approved):

pom.xml
```diff
@@ -240,9 +240,9 @@
                             <goal>enforcer</goal>
                         </goals>
                         <configuration>
-                            <referenceVersion>5.24ea0</referenceVersion>
+                            <referenceVersion>2.25ea0</referenceVersion>
```
tgd, line 243: Why is this line changing?

yevgenp: Sorry, that's a mistake

tgd: No problem at all

PR #1473, opened 2023-11-13
Ensure DiskSpaceMonitor is initialised on each Queue instantiation

tgd:


rogersimmons (approved):

JerryShea (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1004,8 +1006,6 @@ public SingleChronicleQueueStore acquire(int cycle, CreateStrategy createStrateg
                     return null;
                 }
 
-                if (createStrategy != CreateStrategy.READ_ONLY)
```
JerryShea, line 1007: agree it makes much more sense to initialise this in queue ctor rather than here

JerryShea (approved):

PR #1478, opened 2023-11-22
Add NT version-index lock

tgd:


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -583,6 +584,16 @@ public LongValue indexVersionForId(@NotNull String id) {
         return this.metaStore.doWithExclusiveLock((ts) -> ts.acquireValueFor(String.format(INDEX_VERSION_FORMAT, id), -1L));
     }
 
+    @NotNull
+    public TableStoreWriteLock versionIndexLockForId(@NotNull String id) {
+        return new TableStoreWriteLock(
+                metaStore,
+                pauserSupplier,
+                timeoutMS * 3 / 2,
```
tgd, line 592: These arguments are passed to the append lock and have been copied here for now. A further review of the timeout and pauser supplier is needed before merge

rogersimmons (approved):

pom.xml
```diff
@@ -27,7 +27,7 @@
 
     <artifactId>chronicle-queue</artifactId>
 
-    <version>5.25ea1-SNAPSHOT</version>
+    <version>5.25ea1-nt_locking-SNAPSHOT</version>
```
tgd: Added for CI, needs removal prior to merge.

nicktindall (approved):

PR #1479, opened 2023-11-23
Remove deprecated code scheduled for removal in x.25

nicktindall:
Hide whitespace for easier diff


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -650,78 +619,80 @@ public long exactExcerptsInCycle(int cycle) {
...
-        if (fromIndex > toIndex) {
-            long temp = fromIndex;
-            fromIndex = toIndex;
-            toIndex = temp;
-        }
-
-        // if the are the same
-        if (fromIndex == toIndex)
-            return 0;
+        try (ExcerptTailer tailer = createTailer()) {
```
nicktindall, line 622: This is less garbage-y than it was, but it's garbage-y. I wonder if we should provide an overload that takes a Tailer as a zero-garbage option, and have this one call it with a tailer it creates and closes.

rogersimmons: This feels like a reasonable use case for a thread-local which is lazy instantiated then reset and used when needed.

JerryShea: Thread locals have caused all kinds of leakage problems and we have removed them elsewhere as customers ended up surprised. I think the overload is more intentional, even though it makes the API bigger

JerryShea (approved):
Approved. The countExcerpts question can be resolved later IMO

tgd (approved):

PR #1488, opened 2024-01-03
Deflake NormaliseEOFsTest

tgd:
- Take defensive copy of exceptionMap to prevent ConcurrentModificationException that causes test flakiness
- Move test data to OS.getTarget() directory
- Clean up test data after run

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
@@ -1,6 +1,7 @@
 package net.openhft.chronicle.queue.impl.single;
 
 import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.OS;
 import net.openhft.chronicle.core.io.IOTools;
 import net.openhft.chronicle.core.onoes.ExceptionKey;
 import net.openhft.chronicle.core.onoes.LogLevel;
@@ -15,6 +16,9 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.File;
+import java.nio.file.Paths;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
```diff
@@ -65,7 +74,8 @@ public void normaliseShouldResumeFromPreviousNormalisation() {
                 excerptAppender.normaliseEOFs();
             }
             final Pattern logPattern = Pattern.compile("Normalising from cycle (\\d+)");
-            final List<Integer> startIndices = exceptionMap.keySet().stream()
+            // Note a defensive copy of exceptionMap is used as this code has yielded concurrent modification exceptions leading to flakiness under load
```
nicktindall, line 77: Where does the modification come from? Is the stream processing throwing an exception that gets logged? Or some background processing? I wouldn't have thought there was any concurrency in this test?
Perhaps there's a better way to test this?

nicktindall (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/NormaliseEOFsTest.java
tgd: Hi Nick - I believe that the modification is due to `DiskSpaceMonitor` emitting an info log line during the test runtime. I will check now, separately why that DiskMonitor is running during this test and fix the root cause.

I'll send an update on this PR shortly

tgd: Hi @nicktindall dug into this a little more. There are a number of Chronicle threads running in the background of this test that may be mutating the state of this map - this state bleeds from other tests. Dumping the thread names prior to test execution, we see:

```
-------- Dumping thread names
surefire-forkedjvm-command-thread
Signal Dispatcher
process reaper
surefire-forkedjvm-stream-flusher
Reference Handler
main/disk~space~checker
Finalizer
background~resource~releaser
main
-------- Finished dumping thread names
```
Unfortunately I have still not been able to pin down the exact source of the mutation locally (can't get the test to fail). 

I reckon we merge this in for now and see if it reoccurs. I would like to fix it fully when we have some more info.

As an aside we should consider what kind of additional thread and state clean up we should add after each test execution.

nicktindall: Yep best to merge if it addresses the flakiness. We can consider more robust ways to test later.

PR #1491, opened 2024-01-03
Deflake EmptyRollCycleTest on windows

tgd:


peter-lawrey (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/EmptyRollCycleTest.java
```diff
@@ -137,6 +139,11 @@ private void createQueueWithEmptyRollCycleAtEnd() throws IOException {
 
         // Delete the last roll cycle
         final Path lastRollCycle = dataDirectory.resolve(EMPTY_ROLL_CYCLE_NAME);
+        // This test was flaky on windows some small percentage of time due to some lingering file handles
+        // Ensure everything is fully cleaned up before deleting the file
+        if (OS.isWindows()) {
```
peter-lawrey: I would drop the OS check. Should be ok to run everywhere

tgd: Agreed - that's done

PR #1494, opened 2024-01-05
Backport toEnd optimisations to 2.23

tgd:


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -1017,6 +1015,17 @@ public ExcerptTailer originalToEnd() {
...
 
+    private boolean originalToEndLoopCondition(long approximateLastIndex, long index) {
+        if (direction == FORWARD) {
+            return true;
+        } else if (direction == BACKWARD) {
+            // Do not let index run past the approximate last index
+            return index < approximateLastIndex;
+        } else {
+            return true;
+        }
```
yevgenp, line 1026: The whole method can be simplified as `if(condition) return smth`. i.e. dropping all `else`.

src/test/java/net/openhft/chronicle/queue/impl/single/stress/backwardstailer/BackwardsTailerBoundaryTest.java
```diff
@@ -0,0 +1,105 @@
...
+    private final RollCycle rollCycle;
+
+    public BackwardsTailerBoundaryTest(RollCycle rollCycle) {
+        this.rollCycle = rollCycle;
+    }
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection<Object[]> data() {
+        final List<Object[]> data = new ArrayList<>();
+        data.add(new Object[]{RollCycles.TEST4_DAILY});
```
yevgenp, line 39: Not so much parameters :) 
I'd recommend to use a JUnit5 parameterized test (slowly migrating of Junit4)

tgd: :-) Junit5 would have been nicer!

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
tgd: I agree. I would usually change this, however the code is already integrated into `develop` in this form and we are now backporting to the 2.23 release. I'll be revisiting the code in develop so will address there.

PR #1495, opened 2024-01-08
fixed deprecated issues in chronicle services

RobAustin:
see also - https://github.com/ChronicleEnterprise/Chronicle-Services-Enterprise/pull/596 and https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/pull/643

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
+++ b/src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
@@ -169,9 +169,9 @@ public interface ChronicleQueue extends Closeable {
      *
      * @param id unique id for a tailer which uses to track where it was up to
      * @return a new ExcerptTailer for this ChronicleQueue with the given unique {@code id}
-     * @see #createTailer()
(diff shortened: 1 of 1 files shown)


tgd (approved):
I am happy to approve (see email) however important we get Jerry + Nick's feedback prior to merge as they proposed the initial deprecation (I think)

src/main/java/net/openhft/chronicle/queue/ChronicleQueue.java
```diff
@@ -196,9 +196,7 @@ default LongValue indexForId(String id) {
      * this method every time an appender is needed.
      *
      * @return Returns a ExcerptAppender for this ChronicleQueue that is local to the current Thread
-     * @deprecated It is recommended to use {@link #createAppender()} instead
      */
-    @Deprecated(/* To be removed in x.27 */)
```
JerryShea, line 201: needs broader discussion before un-deprecating

peter-lawrey: We decided that making services fail on deprecation was more important.

PR #1496, opened 2024-01-09
added threadLocalAppender

RobAustin:
This PR puts back the deprecation of net.openhft.chronicle.queue.ChronicleQueue#acquireAppender and instead introduces a utility class.

relates to https://github.com/ChronicleEnterprise/Chronicle-Services-Enterprise/pull/599

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/ThreadLocalAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/ThreadLocalAppender.java
@@ -1,0 +1,16 @@
+package net.openhft.chronicle.queue.impl.single;
+
+import net.openhft.chronicle.queue.ExcerptAppender;
+
+public class ThreadLocalAppender {
+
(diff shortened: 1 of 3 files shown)


tgd (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/ThreadLocalAppender.java
```diff
@@ -0,0 +1,16 @@
+package net.openhft.chronicle.queue.impl.single;
+
+import net.openhft.chronicle.queue.ExcerptAppender;
+
+public class ThreadLocalAppender {
+
+    /**
+     * A thread local appender for the SingleChronicleQueue, this will give you the same thread local instance per thread.
+     *
```
JerryShea, line 9: we need the javadoc from `acquireAppender` copied in here with all the reasons and caveats that you need to be aware of when using this

PR #1497, opened 2024-01-10
acquireAppender further cleanup

tgd:
acquireAppender will be deprecated in the future, ensure no calls to it remain in the Chronicle-Queue project. Change the signature on ThreadLocalAppender to accept a ChronicleQueue.


RobAustin (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/ThreadLocalAppender.java
```diff
@@ -1,16 +1,39 @@
...
+     * this method every time an appender is needed.
      *
-     * @param queue the SingleChronicleQueue
-     * @return a thread local appender
+     * @return Returns a ExcerptAppender for this ChronicleQueue that is local to the current Thread
+     * @throws IllegalArgumentException if the queue it is passed is not an instance of {@link SingleChronicleQueue}
      */
-    public static ExcerptAppender acquireThreadLocalAppender(SingleChronicleQueue queue) {
-        return queue.acquireThreadLocalAppender(queue);
+    public static ExcerptAppender acquireThreadLocalAppender(ChronicleQueue queue) {
```
tgd, line 33: @JerryShea, @RobAustin makes a good point - there is a case to leave this accepting `SingleChronicleQueue` because this class is tightly bound to the implementation itself, not something that is available on the interface. By accepting `ChronicleQueue` here we may encourage others to pass any queue into this method. Thoughts?

On the flip side you will need to cast to use this method, but maybe that friction is a good thing

JerryShea: We could create an instance of `ThreadLocalAppender` for a queue and have *it* manage the TL (and implement Closeable etc.) and then (I haven't checked this in detail) maybe we could remove the TL from `SingleChronicleQueue` and it would work for all types of queues.

JerryShea (approved):
This is definitely an improvement, but once this dust has settled, we need to have a hard look at where we are using this new class

src/main/java/net/openhft/chronicle/queue/internal/main/InternalBenchmarkMain.java
```diff
@@ -72,7 +74,7 @@ static void benchmark(int messageSize) {
 
         // Pretoucher will only work with Queue Enterprise in the path
         Thread pretoucher = new Thread(() -> {
-            ExcerptAppender appender = queue.acquireAppender();
+            ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: Use `createAppender` instead

src/main/java/net/openhft/chronicle/queue/internal/main/InternalPingPongMain.java
```diff
@@ -96,7 +98,7 @@ static void pingPong(int size) {
             Jvm.pause(100);
 
             final long finish = System.currentTimeMillis() + runtime * 1000L;
-            final ExcerptAppender appender = queue.acquireAppender();
+            final ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: Again could probably use createAppender in the try/finally block above

src/main/java/net/openhft/chronicle/queue/internal/writer/ChronicleWriter.java
```diff
@@ -37,7 +39,7 @@ public class ChronicleWriter {
 
     public void execute() throws IOException {
         try (final ChronicleQueue queue = ChronicleQueue.singleBuilder(this.basePath).build()) {
-            final ExcerptAppender appender = queue.acquireAppender();
+            final ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: use createAppender in try/finally above

src/test/java/net/openhft/chronicle/queue/TailerCloseInParallelTest.java
```diff
@@ -105,7 +106,7 @@ static void doPerfTest(String file, TestWriter<Bytes<?>> writer, TestReader<Byte
             thread.start();
 
             UncheckedBytes<BytesStore> bytes = new UncheckedBytes<>(BytesStore.empty().bytesForRead());
-            try (ExcerptAppender appender = chronicle.acquireAppender()) {
+            try (ExcerptAppender appender = acquireThreadLocalAppender(chronicle)) {
```
nicktindall: Use createAppender instead

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueCloseTest.java
```diff
@@ -42,18 +44,18 @@ public void testTailAfterClose() {
     @Test
     public void reacquireAppenderAfterClose() {
         try (final ChronicleQueue queue = SingleChronicleQueueBuilder.builder(getTmpDir(), WireType.BINARY).build()) {
-            final ExcerptAppender appender = queue.acquireAppender();
+            final ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall, line 50: Use createAppender instead (in try/finally above)

src/test/java/net/openhft/chronicle/queue/impl/single/NotCompleteTest.java
```diff
@@ -125,7 +126,7 @@ public void testInterruptOrExceptionDuringSerialisation() throws InterruptedExce
                 assertFalse(reader.readOne());
 
                 // do an empty write
-                ExcerptAppender appender = queueWriter.acquireAppender();
+                ExcerptAppender appender = acquireThreadLocalAppender(queueWriter);
```
nicktindall: Use createAppender (can do it in try/finally where queues are created)

src/test/java/net/openhft/chronicle/queue/impl/single/StoreAppenderTest.java
```diff
@@ -58,15 +59,15 @@ public void threadDump() {
...
+            assertThrows(UnsupportedOperationException.class, () -> acquireThreadLocalAppender(queue).singleThreadedCheckReset());
         }
     }
 
     @Ignore("net/openhft/chronicle/queue/channel/PublishHandler.java:25")
     @Test
     public void resetUsedByThreadThrowsUnsupportedOperationException() throws IOException {
         try (SingleChronicleQueue queue = SingleChronicleQueueBuilder.single(queueDirectory.newFolder()).build()) {
-            assertThrows(UnsupportedOperationException.class, () -> queue.acquireAppender().singleThreadedCheckReset());
+            assertThrows(UnsupportedOperationException.class, () -> acquireThreadLocalAppender(queue).singleThreadedCheckReset());
```
nicktindall: These tests are kind of redundant now, I don't even know why they're still in there. Please delete. They were put there to try and prevent people passing thread-local appenders between threads, but someone disabled them later on.

src/test/java/net/openhft/chronicle/queue/impl/single/StoreTailerTest.java
```diff
@@ -213,7 +214,7 @@ public void checkAfterWrittenMessageAtIndexMovesToTheCorrectIndex() {
             outputWriter.onEvent("out1");
 
             // Get the index of the last message appended to the output queue
-            long index = outputQueue.acquireAppender().lastIndexAppended();
+            long index = acquireThreadLocalAppender(outputQueue).lastIndexAppended();
```
nicktindall: Use createAppender

src/test/java/net/openhft/chronicle/queue/impl/single/stress/backwardstailer/BackwardsTailerBoundaryTest.java
```diff
@@ -51,7 +52,7 @@ public void verifyConsistency() {
         @NotNull File path = getTmpDir();
         IOTools.deleteDirWithFiles(path);
         try (SingleChronicleQueue queue = createQueue(path, rollCycle);
-             ExcerptAppender appender = queue.acquireAppender();
+             ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: Use createAppender

src/test/java/net/openhft/chronicle/queue/impl/single/stress/backwardstailer/BackwardsTailerJmhState.java
```diff
@@ -31,7 +32,7 @@ public void setup(int numberOfEntries) {
                 .rollCycle(rollCycle)
                 .build();
         log.info("Populating queue with data");
-        try (ExcerptAppender appender = queue.acquireAppender()) {
+        try (ExcerptAppender appender = acquireThreadLocalAppender(queue)) {
```
nicktindall: Use createAppender

src/test/java/net/openhft/chronicle/queue/impl/single/stress/backwardstailer/BackwardsTailerToEndPerfAcceptanceTest.java
```diff
@@ -113,7 +114,7 @@ private void populateQueue(int entriesToWrite, ExcerptAppender appender) {
     private long runTest(int entriesToWrite, TailerDirection tailerDirection, TailerIndexStartPosition tailerIndexStartPosition, RollCycle rollCycle) {
         @NotNull File path = getTmpDir();
         try (SingleChronicleQueue queue = createQueue(path, rollCycle);
-             ExcerptAppender appender = queue.acquireAppender();
+             ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: Use createAppender

src/test/java/net/openhft/chronicle/queue/method/BrokenChainTest.java
```diff
@@ -28,7 +29,7 @@ interface Second {
     public void brokenChainQueue() {
         try (ChronicleQueue queue = ChronicleQueue.single(OS.getTarget() + "/brokernChain-" + System.nanoTime());
              // using createAppender() doesn't work as the chained methods uses acquireAppender()
-             ExcerptAppender appender = queue.acquireAppender();
+             ExcerptAppender appender = acquireThreadLocalAppender(queue);
```
nicktindall: Use createAppender

nicktindall (commented):
There are very few places where using acquireAppender still makes sense, any time you're closing the acquired appender it's a clear sign that a create would make more sense.

This whole change I'm not a fan of, all it does is leave the underlying issue in place but under a different name/clunkier API.

I agree that currently there are places where we still need acquireAppender, but I'd prefer if we used an internal API to acquire those appenders so it's not used by users and it can be easily removed once we address those resource management issues.

Can we please move ThreadLocalAppender to internal package at least? it'd be good to move away from this pattern. Ideally before we cut a release, so we don't have to go through the whole deprecation process again.

nicktindall (changes requested):
At very least can we use createAppender where it makes sense, but ideally I'd love to see ThreadLocalAppender moved to `internal`.

tgd:
Hi @nicktindall @JerryShea - thanks - noted and changed as per your comments.

I did move the ThreadLocalAppender to `internal`, *however* this means that that `SingleChronicleQueue#acquireThreadLocalAppender` method needs to be opened up to public. I think on balance this is worse than leaving it in impl and would vote to leave it where it is. I do agree it would be nicer in internal but opening the other method up as public is not worth it.

Please can you let me know if you strongly disagree?

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueCloseTest.java
tgd: This test requires the thread local unfortunately

src/test/java/net/openhft/chronicle/queue/impl/single/NotCompleteTest.java
```diff
@@ -170,9 +171,10 @@ private String cleanQueueDump(String from) {
     }
 
     private void doWrite(ChronicleQueue queue, BiConsumer<PersonListener, ChronicleQueue> action) {
-        ExcerptAppender appender = queue.acquireAppender();
-        PersonListener proxy = appender.methodWriterBuilder(PersonListener.class).get();
-        action.accept(proxy, queue);
+        try (ExcerptAppender appender =queue.createAppender()) {
+            PersonListener proxy = appender.methodWriterBuilder(PersonListener.class).get();
+            action.accept(proxy, queue);
+        }
```
nicktindall: I think you can use `ChronicleQueue#methodWriter(PersonListener.class)` here and avoid (or at least hide) the use of `xxxAppender` altogether.

src/test/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueCloseTest.java
nicktindall: Ah yes

src/main/java/net/openhft/chronicle/queue/impl/single/ThreadLocalAppender.java
nicktindall: I like the idea @JerryShea, but it would still pose issues for method writers and channels (namely when do we create and close the `ThreadLocalAppender` instance), the fundamental issue is they don't have a lifecycle and are themselves billed as "thread safe".

I believe our goal here should be making this API as obscure and hidden as possible, while reserving the right to rip it out as soon as we fix the stragglers that require it (channels and method writers from what I can see). If we made those things single threaded and added a lifecycle to them we can put this whole mess behind us and let the API consumer do their resource management and thread-localisation if that's what they desire.

nicktindall (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/NotCompleteTest.java
tgd: Nice! That's changed

PR #1502, opened 2024-01-17
Implement unsafe write bytes. Fixes #1501

### LINKED ISSUE #1501: Allow a fast, less safe append-at-index
`StoreTailer#writeBytesInternal` is very paranoid when validating the index it is passed, to ensure it doesn't append at a different index.

Rather than trusting the metadata written to the store file header, it scans through the store to be absolutely certain it's writing at the specified index. This is an expensive operation.

The write could be much faster if we trusted the value in the header, but that comes with risk of validating the specified index against a stale "last index" and performing a write where it should not. In certain circumstances, such as in replication where the sink handler knows it's the only writer to the queue, we can speed things up by using the header value instead of performing the scan.

nicktindall:
Fixes #1501

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/InternalAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/InternalAppender.java
@@ -22,8 +22,46 @@ import net.openhft.chronicle.bytes.BytesStore;
 import net.openhft.chronicle.queue.ExcerptAppender;
 
 /**
- * please don't use this interface as its an internal implementation.
+ * please don't use this interface as it's an internal implementation.
  */
 public interface InternalAppender extends ExcerptAppender {
+
+    /**
+     * Append an excerpt at the specified index, if the index is a valid next index for the queue.
+     * <p>
+     * If the index is:
+     * <dl>
+     *     <dt>Greater than the next valid indices for the queue</dt>
+     *     <dd>An {@link IllegalIndexException} is thrown</dd>
+     *
+     *     <td>Less than or equal to the last index in the queue</td>
+     *     <dd>The method returns without modifying the queue</dd>
+     * </dl>
+     *
+     * @param index index the index to append at
+     * @param bytes bytes the contents of the excerpt to write
+     * @throws IllegalIndexException if the index specified is larger than the valid next indices of the queue
+     */
     void writeBytes(long index, BytesStore bytes);
+
+    /**
+     * Append an excerpt at the specified index, if the index is a valid next index for the queue.
+     * <p>
+     * If the index is:
+     * <dl>
+     *     <dt>Greater than the next valid indices for the queue</dt>
+     *     <dd>An {@link IllegalIndexException} is thrown</dd>
+     *
+     *     <td>Less than or equal to the last index in the queue</td>
+     *     <dd>The method returns without modifying the queue</dd>
+     * </dl>
+     * Note: This form of the method will do a faster, less reliable check to confirm the index specified is
+     * a valid next index for the state of the queue. This should only be used when we know that there have
+     * been no partial writes to the queue previously.
+     *
+     * @param index index the index to append at
+     * @param bytes bytes the contents of the excerpt to write
+     * @throws IllegalIndexException if the index specified is larger than the valid next indices of the queue
+     */
+    void unsafeWriteBytes(long index, BytesStore bytes);
 }
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
@@ -52,6 +52,22 @@ import static net.openhft.chronicle.wire.Wires.*;
 
 class StoreAppender extends AbstractCloseable
         implements ExcerptAppender, ExcerptContext, InternalAppender, MicroTouched {
+
+    /**
+     * Determines behaviour when validating that the index passed to {@link #writeBytesInternal(long, BytesStore, IndexValidationMode)}
+     * is at or beyond the end of the queue.
+     */
+    protected enum IndexValidationMode {
+        /**
+         * We read the queue contents when validating
+         */
+        SAFE,
+        /**
+         * We trust the last index/position header in the store file when validating
+         */
+        UNSAFE
+    }
+
     /**
      * Keep track of where we've normalised EOFs to, so we don't re-do immutable, older cycles every time.
      * This is the key in the table-store where we store that information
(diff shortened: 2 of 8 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/InternalAppender.java
```diff
@@ -26,4 +26,6 @@
  */
 public interface InternalAppender extends ExcerptAppender {
     void writeBytes(long index, BytesStore bytes);
+
+    void unsafeWriteBytes(long index, BytesStore bytes);
```
nicktindall, line 66: Exposed via this interface to make benchmarking easier

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -52,6 +52,22 @@
...
+    protected enum IndexValidationMode {
+        /**
+         * We read the queue contents when validating
+         */
+        SAFE,
+        /**
+         * We trust the last index/position header in the store file when validating
+         */
+        UNSAFE
+    }
```
nicktindall, line 69: Rather than adding another boolean parameter

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -657,14 +717,14 @@ protected void writeBytesInternal(final long index, @NotNull final BytesStore by
...
-        resetPosition(true);
+        resetPosition(indexValidationMode == IndexValidationMode.SAFE);
 
         long headerNumber = wire.headerNumber();
 
         boolean isNextIndex = index == headerNumber + 1;
         if (!isNextIndex) {
             if (index > headerNumber + 1)
-                throw new IllegalStateException("Unable to move to index " + Long.toHexString(index) + " beyond the end of the queue, current: " + Long.toHexString(headerNumber));
+                throw new IndexOutOfBoundsException("Unable to move to index " + Long.toHexString(index) + " beyond the end of the queue, current: " + Long.toHexString(headerNumber));
```
nicktindall: This is potentially controversial.

While I doubt anyone was catching this `IllegalStateException`, throwing `IndexOutOfBoundsException` is an interface change, but arguably a better fit for what's occurred here, and it also allows you to distinguish programmatically between the index failing validation (and no write being performed) and discovering we wrote at the wrong index AFTER the excerpt was written (as is checked below).

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -630,21 +645,66 @@ public void writeBytes(final long index, @NotNull final BytesStore bytes) {
...
+     * @param index               Index to append at
+     * @param bytes               The excerpt bytes
+     * @param metadata            Whether to write the excerpt as metadata, this is legacy and should always be set to false
+     * @param indexValidationMode How stringent to be when validating the append index
+     * @throws IndexOutOfBoundsException when the index specified is not after the end of the queue
+     * @deprecated Use {@link #writeBytesInternal(long, BytesStore, IndexValidationMode)} instead
+     */
+    @SuppressWarnings("DeprecatedIsStillUsed")  // will be collapsed into the caller when deprecated
+    @Deprecated(/* For removal in x.27 */)
+    protected void writeBytesInternal(final long index, @NotNull final BytesStore bytes, boolean metadata, IndexValidationMode indexValidationMode) {
```
nicktindall, line 698: I've deprecated this method in the interest of removing the metadata parameter.

Firstly I don't think it makes any sense to append a metadata excerpt at a specified index because metadata excerpts don't have indices.

Secondly I believe the check on line 740 would always fail after appending a metadata excerpt, as it will not have incremented the `headerNumber` field (see [here](https://github.com/OpenHFT/Chronicle-Wire/blob/2ca5caf357571eccbdf8cbcfee2ae0438f18e6c1/src/main/java/net/openhft/chronicle/wire/AbstractWire.java#L366)).

So unless anyone objects, this change will remove the ability to `writeBytesInternal` a metadata excerpt at a specified index. An ability that arguably doesn't currently exist anyway.

JerryShea: Love it

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
JerryShea: I vote for a new subclass of IllegalStateException ([we use that pattern commonly](https://github.com/search?q=org%3AOpenHFT%20%22extends%20IllegalStateException%22&type=code))

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: Proven. This method never works when `metadata=true`

See https://github.com/OpenHFT/Chronicle-Queue/commit/f4248587d352e9b9d2663ab1665adba6a8fde0f9 on branch [demonstrate_that_writeBytesInternal_does_not_work_for_metadata](https://github.com/OpenHFT/Chronicle-Queue/tree/demonstrate_that_writeBytesInternal_does_not_work_for_metadata)

src/test/java/net/openhft/chronicle/queue/bench/InternalAppenderJLBH.java
```diff
@@ -63,13 +66,18 @@ public void init(JLBH jlbh) {
...
+        } else {
+            appender.writeBytes(index, payload);
+        }
         jlbh.sample(System.nanoTime() - startTimeNS);
         sequenceNumber++;
     }
 
     @Override
     public void complete() {
+        TeamCityHelper.teamCityStatsLastRun(InternalAppenderJLBH.class.getSimpleName() + (UNSAFE_APPENDS ? "-UNSAFE" : "-SAFE"), jlbh, ITERATIONS, System.out);
```
JerryShea, line 80: great - I will close #1503

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
JerryShea: Fantastic

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
nicktindall: It really should be an `IllegalArgumentException` anyway, not an `IllegalStateException`

The second check (after we've done the append) makes sense as an `IllegalStateException`

JerryShea: I hear you. Could we change it to IAE in x.26?

nicktindall: Well it is marked as an internal API, you really need to reach under the hood, or use `InternalAppender` interface to get at it.

nicktindall: The method also says
`<p>Users are advised that the behaviour of this method may change in the future` in the javadoc

JerryShea: Yes fair point - it is an internal method. Lets change it, but I think IAE is safer than IndexOutOfBoundsE as our convention tends to throw IAE, ISE, or subclasses thereof

nicktindall: Yeah cool, will do the specialised IAE

JerryShea (approved):

tgd (approved):
Looking great Nick

src/test/java/net/openhft/chronicle/queue/InternalAppenderWriteBytesTest.java
```diff
@@ -202,74 +226,35 @@ private SingleChronicleQueue createQueue(File tmpDir) {
...
-
-            try (DocumentContext dc = tailer.readingDocument()) {
-                dc.rollbackOnClose();
-            }
-
-            assertEquals(1, tailer.index());
-
-            ((InternalAppender) appender).writeBytes(2, test);
-        }
-    }
```
nicktindall, line 246: I have no idea what this was testing....

PR #1503, opened 2024-01-17
add TeamCityHelper to InternalAppenderJLBH

JerryShea:


nicktindall:
Ah I did that in https://github.com/OpenHFT/Chronicle-Queue/pull/1502, we could wait for that or merge this and I'll rebase on top of it

PR #1510, opened 2024-01-24
Fix for skipped roll cycle due to bad header. Fixes #1509

### LINKED ISSUE #1509: Tailers can skip unacquirable cycles
When reading through a queue, if a cycle is encountered with an incomplete or invalid header, a tailer will pause (for, by default, 10s) waiting for the header to become valid.

The assumption is it has encountered a new roll cycle in the process of being initialised, and the actual wait time will be much shorter.

If a process dies abruptly while initialising a roll cycle it can leave a bad roll cycle that will hold up all readers/appenders for the full 10s, before an appender will go in and rename the file and replace it with a valid one.

[Some work was done recently](https://github.com/OpenHFT/Chronicle-Queue/pull/1368) to better handle missing roll cycles, when a Tailer failed to acquire a cycle it was assumed to be due to the file being missing, so it would skip forward to the next one.

If a Tailer fails to acquire a cycle because it's partly initialised that is indistinguishable from it failing to acquire one because it's missing from disk. In the former case, it's possible that while it was waiting 10s for the header to be complted, that roll cycle was replaced, populated and the following roll cycle was created (most likely in a replication scenario or one with very short roll cycles). In this scenario the tailer would skip over the replacement cycle and miss a bunch of excerpts.

This caused flappiness in UngracefulShutdownTest.

nicktindall:
Fixesd #1509

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -530,16 +530,15 @@ class StoreTailer extends AbstractCloseable
             nextIndex = nextIndexWithinFoundCycle(nextCycle);
         else
             try {
-                int nextCycle0 = this.cycle;
-                while (true) {
-                    nextCycle0 = queue.nextCycle(nextCycle0, direction);
-                    if (nextCycle0 == -1)
-                        return Long.MIN_VALUE;
+                final int nextCycle0 = queue.nextCycle(this.cycle, direction);
+                if (nextCycle0 == -1)
+                    return Long.MIN_VALUE;
 
-                    if (cycle(nextCycle0)) {
-                        nextIndex = nextIndexWithinFoundCycle(nextCycle0);
-                        break;
-                    }
+                if (cycle(nextCycle0)) {
+                    nextIndex = nextIndexWithinFoundCycle(nextCycle0);
+                } else {
+                    queue.refreshDirectoryListing();
+                    return Long.MIN_VALUE;
                 }
 
             } catch (ParseException e) {
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
@@ -126,8 +126,10 @@ class TableDirectoryListing extends AbstractCloseable implements DirectoryListin
             if (!INITIAL_MAX_FILENAME.equals(maxFilename))
                 max = fileNameToCycleFunction.applyAsInt(maxFilename);
 
-            if (currentMin0 == min && currentMax0 == max)
+            if (currentMin0 == min && currentMax0 == max) {
+                modCount.addAtomicValue(1);
                 return;
+            }
 
             minCycleValue.setOrderedValue(min);
             if (maxCycleValue.compareAndSwapValue(currentMax, max)) {
--- a/src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
+++ b/src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
@@ -391,7 +391,7 @@ public class TestDeleteQueueFile extends QueueTestCommon {
                         long lastReadIndex = -5;
                         int currentCycle = -1;
                         while (running.get()) {
-                            try (final DocumentContext documentContext = tailer.readingDocument()) {
+                            try (final DocumentContext documentContext = readingDocumentWithRetries(tailer)) {
                                 if (!documentContext.isPresent()) {
                                     logIterationResult(direction, tailer, cyclesRead, lastReadIndex);
                                     break;
@@ -414,6 +414,17 @@ public class TestDeleteQueueFile extends QueueTestCommon {
                 Jvm.error().on(TestDeleteQueueFile.class, "Error occurred", e);
             }
             Jvm.startup().on(TestDeleteQueueFile.class, "Tailer thread terminated: " + direction);
+        }
+
+        private DocumentContext readingDocumentWithRetries(ExcerptTailer excerptTailer) {
+            DocumentContext documentContext = null;
+            for (int i = 0; i < 2; i++) {
+                documentContext = excerptTailer.readingDocument();
+                if (documentContext.isPresent()) {
+                    break;
+                }
+            }
+            return documentContext;
         }
 
         private OptionalLong lastAvailableIndex() {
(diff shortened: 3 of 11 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -530,16 +530,15 @@ private long nextIndexWithNextAvailableCycle0(final int cycle) {
...
-                    }
+                final int nextCycle0 = queue.nextCycle(this.cycle, direction);
+                if (nextCycle0 == -1)
+                    return Long.MIN_VALUE;
+
+                if (cycle(nextCycle0)) {
+                    nextIndex = nextIndexWithinFoundCycle(nextCycle0);
+                } else {
+                    queue.refreshDirectoryListing();
+                    return Long.MIN_VALUE;
```
nicktindall, line 541: The failure to acquire a cycle here is not necessarily due to the cycle not existing. When it occurs, we have to return that no next cycle was found, and force-refresh the directory listing to cover off the case where it IS a missing file.

src/main/java/net/openhft/chronicle/queue/impl/single/TableDirectoryListing.java
```diff
@@ -126,8 +126,10 @@ public void refresh(final boolean force) {
             if (!INITIAL_MAX_FILENAME.equals(maxFilename))
                 max = fileNameToCycleFunction.applyAsInt(maxFilename);
 
-            if (currentMin0 == min && currentMax0 == max)
+            if (currentMin0 == min && currentMax0 == max) {
+                modCount.addAtomicValue(1);
                 return;
+            }
```
nicktindall, line 132: If someone did a force refresh, even if the min and max cycles didn't change, we should bump the modCount to trigger a rebuild of the cycle tree. This allows us to move past missing files.

src/test/java/net/openhft/chronicle/queue/impl/single/RollCycleTest.java
```diff
@@ -289,7 +289,7 @@ public void newRollCycleIgnored2() throws InterruptedException {
                                 "\"3\"\n" +
                                 "...\n" +
                                 "# 130660 bytes remaining\n",
-                        queue.dump());
+                        queue.dump().replaceAll("listing.modCount: \\d+", "listing.modCount: 9"));
```
nicktindall, line 292: This seems to change slightly due to concurrency/performance or something, just ignore modCount

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -416,6 +416,17 @@ public void run() {
...
+        private DocumentContext readingDocumentWithRetries(ExcerptTailer excerptTailer) {
+            DocumentContext documentContext = null;
+            for (int i = 0; i < 2; i++) {
+                documentContext = excerptTailer.readingDocument();
+                if (documentContext.isPresent()) {
+                    break;
+                }
+            }
+            return documentContext;
+        }
```
nicktindall, line 428: We need to retry here, but that's OK, that's how people will use the API, the point is to prove with repeated calls to readingDocument you can progress through a queue with files being deleted from it.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -530,16 +530,15 @@ private long nextIndexWithNextAvailableCycle0(final int cycle) {
             nextIndex = nextIndexWithinFoundCycle(nextCycle);
         else
             try {
-                int nextCycle0 = this.cycle;
-                while (true) {
```
tgd, line 534: @nicktindall did this loop ever reach a second iteration or always breaks prior to that?

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
```diff
@@ -416,6 +416,17 @@ public void run() {
             Jvm.startup().on(TestDeleteQueueFile.class, "Tailer thread terminated: " + direction);
         }
 
+        private DocumentContext readingDocumentWithRetries(ExcerptTailer excerptTailer) {
+            DocumentContext documentContext = null;
+            for (int i = 0; i < 2; i++) {
```
tgd, line 421: Is there any significance to the number chosen `2` here?

tgd (approved):
Thanks for sorting this Nick. I added a few questions which would be good to understand.

Would like a second pair of eyes from @JerryShea here too if poss.

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
nicktindall: The problem arose when this loop did make it to a second iteration, due to the first cycle being un-acquirable because of a bad header.

one such sequence of events was (two threads, cycle 3 is last cycle, with a bad header)

|tailer|appender|
|--------|------|
| | acquire cycle 3 for write|
|acquire cycle 3 for read|   |
|  | acquire 3 times out waiting for header|
|  | move invalid cycle 3 file|
|  | create/populate cycle 3 |
|  | create/populate cycle 4 |
|acquire 3 times out waiting for header | |
|assume 3 was deleted, skip to 4| | 

on that last row you just  skipped over the entries in (now valid) 3. It's an obscure sequence of events but can happen with very short (e.g. secondly) roll cycles, or in cases where the Appender is a replicator. Both of these things are true in UngracefulShutdownTest.

This change removes the (incorrect) assumption that `cycle(nextCycle0)` only returns false when the file doesn't exist.

src/test/java/net/openhft/chronicle/queue/TestDeleteQueueFile.java
nicktindall: The first call to `readingDocument()` will return a not-present document context when a deleted cycle is encountered, but it will invalidate the cached cycle tree, so a subsequent call will rebuild that from disk and be able to skip forward to the next actually-present cycle. There's probably a small chance that it hits a deleted cycle twice in a row, but maybe we monitor that and see how likely it is.

nicktindall:
> Thanks for sorting this Nick. I added a few questions which would be good to understand.
> 
> Would like a second pair of eyes from @JerryShea here too if poss.

Yes definitely, the more eyes the better on this.

JerryShea:
> Thanks for sorting this Nick. I added a few questions which would be good to understand.
> 
> Would like a second pair of eyes from @JerryShea here too if poss.

LGTM but I raise your second pair of eyes request to @peter-lawrey

JerryShea (approved):

peter-lawrey (approved):
Great investigation

PR #1520, opened 2024-01-31
Remove @NotNull from catch clauses to avoid compilation error.

yevgenp:


tgd (approved):
Context from Yevgen - this fixes compilation issues on Windows that predate the notnull pom changes

PR #1527, opened 2024-02-07
POC consistency check (DO NOT MERGE)

nicktindall:
This implements a prototype approach for detecting and correcting partial updates that can occur as a result of an abrupt termination

Summary

If we acquire a write lock and `writeInProgress` is true it means the last lock holder did not complete the write, we then do a scan of the roll cycle to ensure the `writePositionAndSequence` table store header and index is consistent with the roll cycle state.

If we knew when the write lock was force-acquired we could probably do away with the `writeInProgress` flag, as a lingering write lock is a good indication that a consistency check is required.


nicktindall:
Superseded by https://github.com/OpenHFT/Chronicle-Queue/pull/1537

PR #1530, opened 2024-02-14
Deflake AppenderFileHandleLeakTest. Test tries to ensure items are cl…

tgd:
Caused build-all breakages. Waiting for tests to run.

Please ignore Win+Mac breaks, they are known and have been investigated need to be addressed next month:

```bootstrap class path not set in conjunction with -source 8```

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
@@ -61,7 +61,7 @@ public final class AppenderFileHandleLeakTest extends QueueTestCommon {
     private static final SystemTimeProvider SYSTEM_TIME_PROVIDER = SystemTimeProvider.INSTANCE;
     private static final RollCycle ROLL_CYCLE = TEST_SECONDLY;
     private static final DateTimeFormatter ROLL_CYCLE_FORMATTER = DateTimeFormatter.ofPattern(ROLL_CYCLE.format()).withZone(ZoneId.of("UTC"));
-    private static final int TRIES = 100;
+    private static final int TRIES = 10;
 
     private final ExecutorService threadPool = Executors.newFixedThreadPool(THREAD_COUNT,
             new NamedThreadFactory("test"));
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java [hunk 1]
```diff
@@ -61,7 +61,7 @@ public final class AppenderFileHandleLeakTest extends QueueTestCommon {
     private static final SystemTimeProvider SYSTEM_TIME_PROVIDER = SystemTimeProvider.INSTANCE;
     private static final RollCycle ROLL_CYCLE = TEST_SECONDLY;
     private static final DateTimeFormatter ROLL_CYCLE_FORMATTER = DateTimeFormatter.ofPattern(ROLL_CYCLE.format()).withZone(ZoneId.of("UTC"));
-    private static final int TRIES = 5;
+    private static final int TRIES = 100;
```
yevgenp: Wow, that may seriously slow down the build.

yevgenp (approved):

src/test/java/net/openhft/chronicle/queue/impl/single/AppenderFileHandleLeakTest.java
tgd: Hi Yevgen, it is used here:

```java
        for (int i = 0; i < TRIES; i++) {
            GcControls.waitForGcCycle();
            openQueueFiles = MappedFileUtil.getAllMappedFiles().stream()
                    .filter(str -> str.contains(queuePath.getAbsolutePath()))
                    .collect(Collectors.toList());
            if (openQueueFiles.isEmpty())
                return true;
            Jvm.pause(10);
        }
```
It should slow it down in these assertions to (in this class only) worst case 100 * 10 -> 1000ms -> 1s

tgd: Actually you're right to call this out. `waitForGcCycle` will also wait for up to a second... Let me revise the try count down.

tgd: Going from 5->10 seems to bring the tests relatively stable on Linux for now. Let's go with the 5 over the 100. Thanks for flagging.

PR #1533, opened 2024-02-16
Fix inconsistencies introduced in side-effects of StoreTailer#approximateLastCycle2(int)

### LINKED ISSUE #1532: toEnd doesn't work when the last roll cycle is empty
If the last roll cycle in the queue is empty, `toEnd` is broken.

This is a result of the internal state of the Tailer being made inconsistent by the `approximateLastCycle2` scan.

nicktindall:
Fixes #1532

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
@@ -783,6 +783,7 @@ class StoreTailer extends AbstractCloseable
             if (this.store != wireStore) {
                 releaseStore();
                 this.store = wireStore;
+                this.cycle = lastCycle;
                 resetWires();
             }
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -783,6 +783,7 @@ private long approximateLastCycle2(int lastCycle) throws StreamCorruptedExceptio
             if (this.store != wireStore) {
                 releaseStore();
                 this.store = wireStore;
+                this.cycle = lastCycle;
```
nicktindall, line 786: We used to update the `store` field without making a corresponding change to the `cycle` field.

JerryShea (approved):

JerryShea:
love it!

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
nicktindall: this caused issues for the logic above that says:

` wireStore = (cycle == lastCycle) ? this.store : queue.storeForCycle(lastCycle, queue.epoch(), false, this.store);`

on the subsequent recursion.

tgd:
Great work Nick

PR #1537, opened 2024-02-29
Make queue tolerate partial update to sequence/writePosition headers

### LINKED ISSUE #1538: When a partial update occurs, we sometimes end up thinking the second last excerpt is the last excerpt
We cache the position/sequence pair of the last excerpt in the queue in the store file header.

It is updated non-atomically with an append.

If the program suffers an abnormal termination it can leave the cached position/sequence pair inconsistent with the actual last excerpt in the queue.

When the pair is updated we merely increment the sequence and write the new position, so this inconsistency is persisted until the end of the roll cycle, and could cause off-by-one entries to be written to the index.

nicktindall:
Fixes #1538 

This is much safer by default, and should speed up any times `exactLastSequence` was being used.

There doesn't seem to be much impact on the [performance tests](https://teamcity.chronicle.software/project/Chronicle_ChronicleQueue?projectTab=stats&branch=#all-projects).

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/QueueSystemProperties.java
+++ b/src/main/java/net/openhft/chronicle/queue/QueueSystemProperties.java
@@ -35,7 +35,7 @@ public final class QueueSystemProperties {
      * Default unset value: false
      * Activation values  : "", "yes", or "true"
      */
-    public static final boolean CHECK_INDEX = Jvm.getBoolean("queue.check.index");
+    public static boolean CHECK_INDEX = Jvm.getBoolean("queue.check.index");
 
     /**
      * Name of a system property used to specify the default roll cycle.
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
@@ -38,6 +38,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.io.EOFException;
 import java.io.StreamCorruptedException;
+import java.io.UncheckedIOException;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -577,7 +578,7 @@ class SCQIndexing extends AbstractCloseable implements Indexing, Demarshallable,
         try {
             return linearScanByPosition(wire, position, indexOfNext, lastKnownAddress, inclusive);
         } catch (EOFException e) {
-            throw new IllegalStateException(e);
+            throw new UncheckedIOException(e);
         }
     }
 
@@ -710,12 +711,12 @@ class SCQIndexing extends AbstractCloseable implements Indexing, Demarshallable,
     }
 
     @Override
-    public long lastSequenceNumber(@NotNull ExcerptContext ec, boolean approximate)
+    public long lastSequenceNumber(@NotNull ExcerptContext ec)
             throws StreamCorruptedException {
         throwExceptionIfClosed();
 
         Sequence sequence1 = this.sequence;
-        if (approximate && sequence1 != null) {
+        if (sequence1 != null) {
             for (int i = 0; i < 128; i++) {
 
                 long address = writePosition.getVolatileValue(0);
@@ -726,7 +727,12 @@ class SCQIndexing extends AbstractCloseable implements Indexing, Demarshallable,
                     continue;
                 if (sequence == Sequence.NOT_FOUND)
                     break;
-                return sequence;
+                try {
+                    Wire wireForIndex = ec.wireForIndex();
+                    return wireForIndex == null ? sequence : linearScanByPosition(wireForIndex, Long.MAX_VALUE, sequence, address, true);
+                } catch (EOFException e) {
+                    throw new UncheckedIOException(e);
+                }
             }
         }
 
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
@@ -315,14 +315,25 @@ public class SingleChronicleQueueStore extends AbstractCloseable implements Wire
         return indexing.sequenceForPosition(ec, position, inclusive);
     }
 
+    /**
+     * @deprecated Use {@link #lastSequenceNumber(ExcerptContext)} instead
+     */
+    @Deprecated(/* To be removed in x.26 */)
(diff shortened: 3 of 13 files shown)


src/test/java/net/openhft/chronicle/queue/impl/single/TestTailAfterRoll.java
```diff
@@ -47,6 +48,7 @@ public class TestTailAfterRoll extends QueueTestCommon {
      * (6) write to this queue created in (5)
      * (7) when you now try to read from this queue you will not be able to read back what you have just written in (6)
      */
+    @Ignore
```
nicktindall, line 51: This test appears to put the queue in a corrupt state... it tries to manually write a EOF to the cycle, which results in the following bytes

```
00010110             c0 eb 68 65  6c 6c 6f 20 77 6f 72 6c     ··he llo worl
00010120 64 00 00 00 c0 00 00 00  00 00 00 00 00 00 00 00 d······· ········
```

When an actual EOF on the cycle would result in the following bytes
```
00010110             c0 eb 68 65  6c 6c 6f 20 77 6f 72 6c     ··he llo worl
00010120 64 00 00 00 00 00 00 c0  c3 45 4f 46 00 00 00 00 d······· ·EOF····
```

It seems to stem from [this issue](https://github.com/OpenHFT/Chronicle-Queue/issues/482), perhaps we can find a better way of testing this? I'm not sure I understand the issue though.

We actually throw an exception if someone tries to append after an EOF, so the described behaviour is not consistent with current behaviour.

@RobAustin perhaps you can advise? I'm inclined to just delete the test and move on

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -726,7 +727,12 @@ public long lastSequenceNumber(@NotNull ExcerptContext ec, boolean approximate)
                     continue;
                 if (sequence == Sequence.NOT_FOUND)
                     break;
-                return sequence;
+                try {
+                    Wire wireForIndex = ec.wireForIndex();
+                    return wireForIndex == null ? sequence : linearScanByPosition(wireForIndex, Long.MAX_VALUE, sequence, address, true);
```
nicktindall, line 732: This is the change, when we get a valid writePosition/sequence pair, we do a linear scan to make sure it's accurate. This covers off the case where we suffered a partial update.

It also removes the need to have separate approximate/exact implementations of `lastSequenceNumber`.

The ONE implementation just does the following
- read the writePosition/sequence header
- if it's consistent, do a quick linear scan to ensure it's up-to-date (this linear scan should be very short, at most 2 excerpts)
- if it's not consistent, traverse the index and roll cycle contents to determine the last sequence

In the absence of concurrent writes (when you hold the write lock), this method should always return the exact last sequence number. If concurrent writes are occurring we can't give that guarantee regardless of the type of scan we do because the end can move at any moment.

src/main/java/net/openhft/chronicle/queue/QueueSystemProperties.java
```diff
@@ -35,7 +35,7 @@ private QueueSystemProperties() {
      * Default unset value: false
      * Activation values  : "", "yes", or "true"
      */
-    public static final boolean CHECK_INDEX = Jvm.getBoolean("queue.check.index");
+    public static boolean CHECK_INDEX = Jvm.getBoolean("queue.check.index");
```
nicktindall, line 38: We need to be able to disable this in order for PartialUpdateTest to replicate what happens in "prod"

src/test/java/net/openhft/chronicle/queue/impl/single/IndexingLastSequenceNumberTest.java
```diff
@@ -27,7 +27,7 @@ void singleCycleOneEntryApproximateLookup() throws StreamCorruptedException {
         assertEquals(0, linearScanByPositionCountStart);
         long lastSequenceNumber = indexing.lastSequenceNumber(appender, true);
         assertEquals(0, lastSequenceNumber);
-        assertEquals(0, indexing.linearScanByPositionCount());
+        assertEquals(1, indexing.linearScanByPositionCount());
```
nicktindall, line 30: we now always do a linear scan, but if we have consistent writePosition/sequence headers it should scan at most 2 non-metadata excerpts

src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java
```diff
@@ -315,14 +315,25 @@ public long sequenceForPosition(@NotNull final ExcerptContext ec, final long pos
...
+     */
+    @Deprecated(/* To be removed in x.26 */)
     public long exactLastSequenceNumber(@NotNull ExcerptContext ec) throws StreamCorruptedException {
+        return lastSequenceNumber(ec);
+    }
+
+    public long lastSequenceNumber(@NotNull ExcerptContext ec) throws StreamCorruptedException {
         throwExceptionIfClosedInSetter();
-        return indexing.lastSequenceNumber(ec, false);
+        return indexing.lastSequenceNumber(ec);
```
nicktindall, line 336: There is only "exact" now

src/test/java/net/openhft/chronicle/queue/impl/single/IndexingLastSequenceNumberTest.java
nicktindall: Perhaps we change this to assert that we don't do the scan-from-start?

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
```diff
@@ -1209,7 +1209,7 @@ public boolean readAfterReplicaAcknowledged() {
     public long approximateExcerptsInCycle(int cycle) {
```
JerryShea: should we deprecate both of these and introduce an `excerptsInCycle` method?

JerryShea (approved):
looks great. Need @peter-lawrey 's blessing

src/main/java/net/openhft/chronicle/queue/impl/single/StoreTailer.java
nicktindall: Done

nicktindall:
Has passed build-all-bumped numerous times https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_BumpedBranch?branch=partial_update_bug&mode=builds

tgd (approved):

tgd:
Great work Nick - I have reviewed and specifically run - `LargeCycleFileToEndPerfMain`. No regression between `develop` and this branch.

PR #1540, opened 2024-02-29
SCQB.clone() to use deepCopy()

peter-lawrey:


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -1117,13 +1117,8 @@ public SingleChronicleQueueBuilder forceDirectoryListingRefreshIntervalMs(long f
         return this;
     }
 
-    @Deprecated(/* to be removed in x.26*/)
```
tgd, line 1120: Hi @peter-lawrey what's the context for this change?

JerryShea (changes requested):
I don't think we should do this. As discussed, we need to add more testing regarding deepCopy vs clone differences

peter-lawrey:
This breaks CQE so won't do this.

PR #1543, opened 2024-03-06
Don't trust the cached last sequence header. Fixes #1538 (Backport to 24)

### LINKED ISSUE #1538: When a partial update occurs, we sometimes end up thinking the second last excerpt is the last excerpt

nicktindall:


src/test/java/net/openhft/chronicle/queue/CheckHalfWrittenMsgNotSeenByTailerTest.java
```diff
@@ -79,6 +80,11 @@ private static void writeIncompleteMessage(String arg, boolean exit) throws Inte
         }
     }
 
+    @Before
+    public void setUp() {
+        ignoreException("Forced unlocking `chronicle.write.lock` in lock");
+    }
```
nicktindall, line 86: Not sure why this was necessary, but this test appeared to fail with it occasionally, which seems expected given what it's doing.

JerryShea (approved):

PR #1550, opened 2024-03-20
Add support for WEEKLY rolling on Sunday 00:00 #1549

peter-lawrey:
This supports the case where a client doesn't want the queue to roll all week, and 4 billion entries are more than enough.
The roll file might get large, e.g. 1 TB, but avoids any rolling during the week.
This will roll on Sunday at 00:00:00 UTC unless the epoch has been specified.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/RollCycles.java
+++ b/src/main/java/net/openhft/chronicle/queue/RollCycles.java
@@ -68,7 +68,7 @@ public enum RollCycles implements RollCycle {
     /**
      * 0xffffffff entries per week, indexing every 256th entry, leave as 4K and 256 for historical reasons. Cycle starts Sunday 00:00
      */
-    WEEKLY(/*-----------*/"yyyy'W'ww", 7 * 24 * 60 * 60 * 1000, 4 << 10, 256, Constants.SUNDAY_00_00),
+    WEEKLY(/*-----------*/"YYYY'W'ww", 7 * 24 * 60 * 60 * 1000, 4 << 10, 256, RollCycleArithmetic.SUNDAY_00_00),
 
     // these are kept for historical reasons
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/net/openhft/chronicle/queue/RollCycles.java
```diff
@@ -65,6 +65,10 @@ public enum RollCycles implements RollCycle {
      * 0xffffffff entries per day, indexing every 256th entry, leave as 4K and 256 for historical reasons.
      */
     FAST_DAILY(/*-----------*/"yyyyMMdd'F'", 24 * 60 * 60 * 1000, 4 << 10, 256),
+    /**
+     * 0xffffffff entries per week, indexing every 256th entry, leave as 4K and 256 for historical reasons. Cycle starts Sunday 00:00
+     */
+    WEEKLY(/*-----------*/"YYYY'W'ww", 7 * 24 * 60 * 60 * 1000, 4 << 10, 256, RollCycleArithmetic.SUNDAY_00_00),
```
nicktindall, line 71: Does it provide much value weekly defaulting to Sunday UTC? that could be as late as 2pm Monday for our friends in Kiribati.

I feel like if you're choosing a weekly roll cycle you really should be nominating an epoch as well.

nicktindall: Actually scratch that it would be 2pm Sunday.... this seems reasonable then

nicktindall (approved):

src/main/java/net/openhft/chronicle/queue/RollCycles.java
```diff
@@ -264,6 +273,11 @@ public int lengthInMillis() {
         return this.lengthInMillis;
     }
 
+    @Override
+    public int defaultEpoch() {
```
JerryShea, line 277: why add this?

JerryShea (approved):

tgd (approved):

src/main/java/net/openhft/chronicle/queue/RollCycles.java
peter-lawrey: So a RollCycle could define an implied epoch, needed for WEEKLY rolling.

RobAustin (approved):

PR #1551, opened 2024-03-22
Review deprecated methods, change build to error on deprecation

peter-lawrey:


src/test/java/net/openhft/chronicle/queue/util/StackSampler.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.util.concurrent.locks.LockSupport;
+
+/**
+ * A utility class for sampling the stack traces of a target thread.
+ * This class creates a background daemon thread which periodically
+ * samples the stack trace of the specified thread and stores the
+ * latest snapshot.
+ */
+public class StackSampler {
```
tgd, line 33: Hi Peter is this new or factored from another class? Might be a good candidate for core

tgd (approved):

src/test/java/net/openhft/chronicle/queue/util/StackSampler.java
peter-lawrey: It was in core, but it was only used in one queue test

PR #1554, opened 2024-04-01
X.26

yevgenp:


docs/How_it_works.adoc
```diff
@@ -50,7 +50,7 @@ import net.openhft.chronicle.queue.impl.single.SingleChronicleQueueBuilder;
 
 // will write the .cq4 file to working directory
 SingleChronicleQueue queue = SingleChronicleQueueBuilder.builder().path(".").build();
-ExcerptAppender appender = queue.acquireAppender();
+ExcerptAppender appender = queue.createAppender();
```
tgd: Example should ideally use try-with-resources to close the appender

src/main/java/net/openhft/chronicle/queue/reader/ChronicleHistoryReader.java
```diff
@@ -160,7 +160,7 @@ public Map<String, Histogram> readChronicle() {
         final WireParselet parselet = parselet();
         final FieldNumberParselet fieldNumberParselet = (methodId, wire) -> parselet.accept(methodIdToName(methodId), wire.read());
         final MessageHistory prev = MessageHistory.get();
-        MessageHistory.set(new VanillaMessageHistory());
+        MessageHistory.emptyHistory();
```
tgd, line 163: Hi Peter - does this need to change back?

tgd (changes requested):
Hi Peter - 2 comments from me - apart from that looks good.

src/main/java/net/openhft/chronicle/queue/reader/ChronicleHistoryReader.java
peter-lawrey: It could be changed back but this is much cleaner

docs/How_it_works.adoc
peter-lawrey: done

src/main/java/net/openhft/chronicle/queue/reader/ChronicleHistoryReader.java
tgd: You're right - I've seen the implementation of `emptyHistory` now - makes sense

tgd (approved):

JerryShea (changes requested):
This needs to be rebased on lastest develop

tgd (approved):

peter-lawrey (approved):

Issue #1555, opened 2024-04-05
Improve the linearScan warning, and make it off by default


peter-lawrey:
Reduce spurious log messages and improve the quality of information when logging does occur. With this change, the linearScan warning is disabled by default - it can be re-enabled with the `chronicle.queue.report.linear.scan.latency` system property, and the threshold for reporting can be changed by setting `linear.scan.warn.ns` - the default is 100Kns.

JerryShea:
Closing this as I believe completed by #1556

hft-team-city:
Released in [Chronicle-Queue-5.24.37](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.24.37), [BOM-2.24.135](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24.135)

hft-team-city:
Released in [Chronicle-Queue-5.24.37](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.24.37), [BOM-2.24.136](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24.136)

PR #1556, opened 2024-04-05
Improve linearSCan warning and make it off by default #1555

peter-lawrey:


src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -348,25 +349,30 @@ private ScanResult linearScan(@NotNull final Wire wire,
...
-        }
+        if (end < start + 100_000)
+            return;
+
+        doPrintLinerScanTime(toIndex, fromKnownIndex, start, desc, end);
     }
 
-    private boolean printLinearScanTime(long toIndex, long fromKnownIndex, long start, long end,
-                                        String desc) {
+    private void doPrintLinerScanTime(long toIndex, long fromKnownIndex, long start, String desc, long end) {
```
JerryShea: Liner should be Linear

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -79,6 +79,7 @@ class SCQIndexing extends AbstractCloseable implements Indexing, Demarshallable,
     int linearScanCount;
     int linearScanByPositionCount;
     Collection<Closeable> closeables = new ArrayList<>();
+    private long lastScanedIndex = -1;
```
tgd: `lastScanedIndex` -> `lastScannedIndex`

tgd (changes requested):
Thanks Peter - just two minor spelling fixes needed

src/main/java/net/openhft/chronicle/queue/impl/single/SCQIndexing.java
```diff
@@ -348,25 +349,30 @@ private ScanResult linearScan(@NotNull final Wire wire,
...
+        // still warming up?
+        if (toIndex <= 1)
+            return;
+
+        // took too long to scan?
         long end = System.nanoTime();
-        if (end > start + 100_000) {
-            printLinearScanTime(toIndex, fromKnownIndex, start, end, "linearScan by index");
-        }
+        if (end < start + 100_000)
```
nicktindall: Should `100_000` be a constant?

perhaps

```java
private static final int LINEAR_SCAN_WARN_THRESHOLD_NS = TimeUnit.MICROSECONDS.toNanos(100);
```

tgd (approved):

PR #1558, opened 2024-04-11
Added fail on warning lint for Javadoc, and rawtypes fixes for Bytes

peter-lawrey:


src/main/java/net/openhft/chronicle/queue/impl/single/RollCycleEncodeSequence.java
```diff
@@ -21,9 +21,8 @@
 import net.openhft.chronicle.core.Maths;
 import net.openhft.chronicle.core.values.LongValue;
 import net.openhft.chronicle.core.values.TwoLongValue;
-import net.openhft.chronicle.wire.Sequence;
```
yevgenp, line 24: Is it necessary? It was more concise.

yevgenp (approved):

src/main/java/net/openhft/chronicle/queue/impl/single/RollCycleEncodeSequence.java
peter-lawrey: Unfortunately yes. You can't put a `@SuppressWarning("deprecation")` on an import, so you need to use the fully qualified name. 
BTW This PR un-deprecates it anyway... https://github.com/OpenHFT/Chronicle-Wire/pull/871

PR #1560, opened 2024-04-12
Bring file contents into line with other antora projects

george-ball:
one of the antora header files was out of step with other antora builds, it has been patched into line


george-ball:
Closing cos need to merge into develop

PR #1569, opened 2024-04-24
disable log.warn if unlocking appendLock of dead process

JerryShea:
Closes https://github.com/ChronicleEnterprise/Chronicle-Queue-Enterprise/issues/706

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
@@ -1192,7 +1192,7 @@ public class SingleChronicleQueueBuilder extends SelfDescribingMarshallable impl
     }
 
     public WriteLock appendLock() {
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -1192,7 +1192,7 @@ public SingleChronicleQueueBuilder setAllNullFields(@Nullable SingleChronicleQue
     }
 
     public WriteLock appendLock() {
-        return readOnly() ? WriteLock.NO_OP : new TableStoreWriteLock(metaStore, pauserSupplier(), timeoutMS() * 3 / 2, TableStoreWriteLock.APPEND_LOCK_KEY);
+        return readOnly() ? WriteLock.NO_OP : new AppendLock(metaStore, pauserSupplier(), timeoutMS() * 3 / 2, TableStoreWriteLock.APPEND_LOCK_KEY);
```
alamar: Perhaps the constant should be taken from AppendLock as well?

JerryShea: Great idea. Done

nicktindall (approved):

PR #1583, opened 2024-06-03
Tidy benchmark

JerryShea:


david-ry4n (approved):
Looks good. I think I've already seen these three lines elsewhere.. chance for static method somewhere?

JerryShea:
That's a good idea - they are duplicated and could/should be in a function called something like `configureForPerformanceBenchmarks`

PR #1584, opened 2024-06-06
Remove old check.thread.safety property

david-ry4n:
The check.thread.saftety property has not been used for a while. Removing references to it.


src/test/java/net/openhft/chronicle/queue/bench/ByteArrayJLBHBenchmark.java
```diff
@@ -33,7 +33,6 @@ public class ByteArrayJLBHBenchmark implements JLBHTask {
     static {
         System.setProperty("disable.thread.safety", "true");
         System.setProperty("jvm.resource.tracing", "false");
-        System.setProperty("check.thread.safety", "false");
```
JerryShea, line 36: haha I only just copy/pasted that!

JerryShea (approved):

PR #1596, opened 2024-07-02
Improve linearSCan warning and make it off by default #1555 (#1556)

james-mcsherry:
* Improve linearSCan warning and make it off by default #1555

* Fix typos


tgd (approved):
Approved - thanks James

PR #1601, opened 2024-07-26
WIP Do not merge. Feature/index race 2338 repro

david-ry4n:
Remove an assertion that allows simulating a potential race condition.


JerryShea:
Thanks for this @david-ry4n - this has now been folded in by @james-mcsherry

PR #1608, opened 2024-08-02
Add method to provide PID for open file

chroniclekevinpowe:
Added method to InternalFileUtil providing map of full file path to PID holding file open.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
+++ b/src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
@@ -97,7 +97,7 @@ public final class InternalFileUtil {
 
         final Stream.Builder<File> builder = Stream.builder();
         try {
-            final Set<String> allOpenFiles = getAllOpenFiles();
+            final Map<String, String> allOpenFiles = getAllOpenFiles();
             for (File file : sortedInitialCandidates) {
                 // If one file is not closed, discard it and the rest in the sequence
                 if (state(file, allOpenFiles) != FileState.CLOSED) break;
@@ -150,16 +150,16 @@ public final class InternalFileUtil {
      * is returned.
      *
      * @param file         to check
-     * @param allOpenFiles The set of all open files retrieve from {@link #getAllOpenFiles()}
+     * @param allOpenFiles The map of all open files retrieve from {@link #getAllOpenFiles()}
      * @return FileState if the given {@code file } is used by any process
      * @throws UnsupportedOperationException if this operation is not
      *                                       supported for the current platform (e.g. Windows).
      */
-    public static FileState state(@NotNull File file, Set<String> allOpenFiles) {
+    public static FileState state(@NotNull File file, Map<String, String> allOpenFiles) {
         assertOsSupported();
         if (!file.exists()) return FileState.NON_EXISTENT;
         final String absolutePath = file.getAbsolutePath();
-        return allOpenFiles.contains(absolutePath)
+        return allOpenFiles.keySet().contains(absolutePath)
                 ? FileState.OPEN
                 : FileState.CLOSED;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
```diff
@@ -254,4 +268,25 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
             return FileVisitResult.CONTINUE;
         }
     }
+
+    private static class ProcFdWalkerWithPid extends ProcFdWalker {
```
benbonavia: would it be simpler to add a method to ProcFdWalker to do something with the file e? and then this class can just implement the action method instead of overriding the whole method?

src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
```diff
@@ -254,4 +268,25 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
...
+
+        private final Map<String, String> openFilesWithPid = new HashMap<>();
+
+        @Override
+        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
+            if (file.toAbsolutePath().toString().matches("/proc/\\d+/fd/\\d+")) {
+                try {
+                    final String e = file.toRealPath().toAbsolutePath().toString();
+                    final String pid = file.getName(1).toString(); // pid holding file open
```
benbonavia: referencing folder by this hardcoded number worries me, if there's an option to save pids elsewhere then it breaks, maybe a more dynamic way of getting the file would be better?

src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
chroniclekevinpowe: Good point - ta!

If the parent class is deprecated it'll make the actioning the deprecation a little more complex, but cleaner in the meantime.

src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
chroniclekevinpowe: Not sure I agree there - it's no more hard-coded than the regex to check the file's naming structure. Are there Linux variants that use a different pathing structure for `/proc`?

If we expanded getAllOpenFIles to more operating systems (which would be keen) the methods used to get information start to vary significantly - on MacOS for example, we're parsing command-line output from lsof.

benbonavia: can you use `file.getFileName().toString()`?

src/test/java/net/openhft/chronicle/queue/util/FileUtilTest.java
```diff
@@ -187,4 +189,25 @@ private <T> void assertSorted(List<T> list, Comparator<T> comparator) {
     protected SingleChronicleQueueBuilder builder(@NotNull File file, @NotNull WireType wireType) {
         return SingleChronicleQueueBuilder.builder(file, wireType).rollCycle(TEST4_DAILY).testBlockSize();
     }
+
+    @Test
+    public void testOpenFilesWithPid() throws IOException {
+        // WIP test
```
JerryShea: can remove this comment?

chroniclekevinpowe: Thank you - done!

src/main/java/net/openhft/chronicle/queue/internal/util/InternalFileUtil.java
```diff
@@ -219,14 +220,16 @@ public static Set<String> getAllOpenFiles() throws IOException {
...
+        private final static int PID_PATH_INDEX = 1; // where is the pid for process holding file open represented in path?
+        private final Map<String, String> openFiles = new HashMap<>();
 
         @Override
         public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
             if (file.toAbsolutePath().toString().matches("/proc/\\d+/fd/\\d+")) {
                 try {
                     final String e = file.toRealPath().toAbsolutePath().toString();
-                    openFiles.add(e);
+                    final String pid = file.getName(PID_PATH_INDEX).toString(); // pid holding file open
```
benbonavia, line 231: this is a much better change!

JerryShea (approved):

benbonavia (approved):

PR #1614, opened 2024-09-18
Improve TableStoreWriteLock 'locked by me' message

chroniclekevinpowe:
Thread ID stored when lock for queue is acquired, and output when attempt to acquire lock by current process fails.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
@@ -17,7 +17,9 @@
  */
 package net.openhft.chronicle.queue.impl.single;
 
+import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.StackTrace;
+import net.openhft.chronicle.core.io.Closeable;
 import net.openhft.chronicle.core.threads.InterruptedRuntimeException;
 import net.openhft.chronicle.queue.impl.TableStore;
 import net.openhft.chronicle.queue.impl.table.AbstractTSQueueLock;
@@ -30,7 +32,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 
-import static net.openhft.chronicle.assertions.AssertUtil.SKIP_ASSERTIONS;
 import static net.openhft.chronicle.core.Jvm.warn;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -66,13 +73,18 @@ public void lock() {
...
         try {
             currentLockValue = lock.getVolatileValue();
+            currentLockTidValue = lockTid.getVolatileValue();
+
             while (!lock.compareAndSwapValue(UNLOCKED, PID)) {
                 currentLockValue = lockGetCurrentLockValue(tlPauser);
             }
-
+            // Set tid explicitly here
+            lockedByThread = Thread.currentThread();
```
JerryShea: `Thread.currentThread()` can be slow - maybe we need to hide this behind a feature property. Maybe we can even piggy back on one of the existing ones.

JerryShea: actually, as discussed, lets re-use existing lockedByThread and enable with ~~jvm.resource.tracing~~ a new property

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -90,10 +93,11 @@ private long lockGetCurrentLockValue(TimingPauser tlPauser) throws TimeoutExcept
         return lock.getVolatileValue();
     }
 
-    private void lockAssertPostConditions() {
-        //noinspection ConstantConditions,AssertWithSideEffects
-        assert SKIP_ASSERTIONS ||
-                ((lockedByThread = Thread.currentThread()) != null && (lockedHere = new StackTrace()) != null);
+    private void checkStoreLockThread() {
+        if (storeLockThread) {
```
JerryShea, line 97: cool

src/main/java/net/openhft/chronicle/queue/impl/single/TableStoreWriteLock.java
```diff
@@ -123,9 +127,10 @@ private String lockHandleTimeoutExCreateWarningMessage(String lockedBy) {
 
     @NotNull
     protected String getLockedBy(long value) {
+        String threadId = lockedByThread == null ? "unknown - " + STORE_LOCK_THREAD + " not set" : Long.toString(lockedByThread.getId());
```
JerryShea, line 130: excellent

JerryShea (approved):

PR #1648, opened 2025-01-10
Suppress display of index in ChronicleReader for empty messages

JerryShea:


benbonavia (approved):

JerryShea:
@benbonavia this will need to be backported

Issue #1704, opened 2026-04-30
Reduce excessive linear scans in lastIndex() hot path


peter-lawrey:
lastIndex() still calls into linear scan logic far more often than expected under concurrent writer/reader load, even
  after the recent sequenceForPosition(Long.MAX_VALUE) fast-path work reduced brute-force indexed-anchor fall-throughs.

#  Example stress output:

  LastIndexFastPathStress: cores=32 writers=4 readers=64 delayNs=0 writes=78766646 lastIndexCalls=314356 retries=9
  retries/call=0.000029 avgLatency=12459680ns maxLatency=2971.10ms

  The retry/fall-through counters suggest the original brute-force path is mostly addressed, but the perf breakdown still
  shows many calls through linearScanByPosition, especially the lastSequenceNumber tail-check path. This means
  lastIndex() is still doing repeated scan work for what should usually be a cheap “latest written sequence” lookup.

#  Areas to investigate:

  - Reduce how often lastIndex() / toEnd() reaches linearScanByPosition.
  - Avoid repeated scan work when writePosition and sequence metadata are already consistent.
  - Consider a dedicated lastIndex/tail lookup path instead of flowing through the broader tailer movement logic.
  - Review hot loops for avoidable interface dispatch and improve JIT inlining opportunities:
      - cache repeated wireForIndex(), bytes(), store(), rollCycle() lookups locally;
      - prefer concrete/internal types in hot private methods where API boundaries are not required;
      - split special Long.MAX_VALUE paths from generic position lookup paths where that removes polymorphic calls or
        branchy logic.
  - Keep semantics around partial writes, metadata documents, and recovery unchanged.

 # Acceptance criteria:

  - Add or keep a stress/regression test that reports scan-source counters.
  - sequenceForPosition brute-force fall-throughs remain near zero under contention.
  - lastIndex() performs substantially fewer linearScanByPosition tail-check calls under the stress test.
  - No public API changes unless explicitly justified.
  - Relevant queue/indexing tests continue to pass.

PR #1706, opened 2026-05-01
Preserve alias safety without rescanning sparse queue tails

peter-lawrey:
The PR preserves alias safety while avoiding unnecessary tail scans. A matching encoded tracker fragment is not treated as an exact full position/sequence pair. The live-codec alias bound and stable position rereads remain, and wrapped positions can use full pairs established by committed writes or independently anchored scans. Otherwise lookup uses persisted indices.

The cache is per physical mapped file and thread, with weak keys. Store reacquisition can reuse a proved pair; a replacement file or a new mapping/thread cannot inherit one. The previously unguarded finite-position, `moveToIndex()` and forward `toEnd()` shortcuts also reject ambiguous tracker pairs. Metadata does not advance the data sequence, and committed suffixes remain part of tail recovery. No file-format change, record-size exemption, one-period probe or new publication protocol is introduced.

The independent audit found and fixed two additional performance regressions at `f51ea463c`:

- After another writer appends 5,000 records, an older reader cache hid newer sparse-index entries. The lookup now prefers newer persisted progress: 5,002 header reads become 906 (within the 1,024-record interval), then two on the next lookup. The public forward reader follows the same rule.
- Warm forward `toEnd()` always fell back above the alias bound. It now scans from a usable full pair and retains the last committed data position, including when metadata follows it. Instrumented header work falls from six operations to three. In the controlled API diagnostic, HUGE_DAILY/HUGE_DAILY_XSPARSE forward-end batch means fall from about 124-127 ns/op to 46-49 ns/op; base is about 40 ns/op.

Six new audit cases pass unchanged on develop `129e9ae5`; four fail on reviewed head `f51ea463c`; all pass with the fixes. They cover writer bursts, public end positioning, suffix/metadata handling, rollback, and a concurrent writer plus independently mapped reader that checks all 5,000 physical record values/indices while querying counts, end positions and finite positions. Original real-codec alias, geometry, backward-roll, retry, suffix and restricted-limit regressions remain. Retained production choices have `//!` comments naming discriminating tests.

Validation of production `ff758d392b3b3fb0405ae69fa08e1388d7572597`:

- Clean full Maven verification on OpenJDK 21.0.12: 1,197 reported tests, zero failures/errors, 52 skipped; licence and binary compatibility pass.
- Clean focused verification on Java 8u502 and actual 32-bit Java 17.0.20: 83 passing tests on each, no skips.
- Nine additional targeted, compiled mutations are killed by their selected tests. A further clean 83-test verification passes after restoration. Surefire reruns are disabled.
- All 35 explicit production test-method references resolve. Exact source hashes, mutation replacements/results, benchmark source, JVM arguments and raw output are included.

The broader diagnostic compares document writes, direct byte writes, forward end, last index, counts and near-tail seeks on FAST_DAILY, HUGE_DAILY and HUGE_DAILY_XSPARSE across base, reviewed and fixed revisions. Writes and warm seeks/end operations retain zero steady-state calling-thread allocation. Counts/lastIndex retain roughly 88 bytes of additional allocation per call; default forward end measures about 6 ns above base in this run. The report documents these costs and measurement limits instead of claiming universal improvement. An apparent default-seek slowdown disappeared with adequate warmup and synchronous compilation.

[Independent audit and current results](https://github.com/OpenHFT/Chronicle-Queue/blob/issues/1704-rewrite/docs/pr-1706-independent-audit.adoc) · [Reproduction and raw evidence](https://github.com/OpenHFT/Chronicle-Queue/tree/issues/1704-rewrite/docs/pr-1706-audit-evidence) · [Earlier review and evidence](https://github.com/OpenHFT/Chronicle-Queue/blob/issues/1704-rewrite/docs/pr-1706-review-findings.adoc)

The findings branch and develop were integrated without rewriting published history. The earlier 23-mutation and JLBH results remain attributed to their exact earlier implementation, rather than relabelled as runs against this head. Issue #1704 remains open for its scan-source counters/concurrent performance criteria and initially cold mappings. PR #1705 remains untouched. Fresh CI must independently validate this updated head.

### DIFF
--- a/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
+++ b/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
@@ -69,6 +69,16 @@ public class SingleChronicleQueueBuilder extends SelfDescribingMarshallable impl
     private static final Constructor<?> ENTERPRISE_QUEUE_CONSTRUCTOR;
     private static final WireStoreFactory storeFactory = SingleChronicleQueueBuilder::createStore;
     private static final Supplier<TimingPauser> TIMING_PAUSER_SUPPLIER = DefaultPauserSupplier.INSTANCE;
+    /**
+     * Maximum park duration (ms) for the writeLock/appendLock pauser.
+     */
+    private static final int WRITE_LOCK_PAUSER_MAX_MS = Jvm.getInteger("chronicle.queue.writeLockPauserMaxMs", 5);
+    /**
+     * Pauser supplier used by {@link #writeLock()} / {@link #appendLock()}. Uses
+     * {@link Pauser#balancedUpToMillis(int)}
+     */
+    private static final Supplier<TimingPauser> WRITE_LOCK_PAUSER_SUPPLIER =
+            () -> Pauser.balancedUpToMillis(WRITE_LOCK_PAUSER_MAX_MS);
 
     static {
         CLASS_ALIASES.addAlias(WireType.class);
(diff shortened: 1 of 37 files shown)


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueBuilder.java
```diff
@@ -711,7 +721,9 @@ public SingleChronicleQueueBuilder createAppenderConditionCreator(Function<Singl
      */
     @NotNull
     WriteLock writeLock() {
-        return readOnly() ? new ReadOnlyWriteLock() : new TableStoreWriteLock(metaStore, pauserSupplier(), timeoutMS() * 3 / 2);
```
tgd, line 714: Consider whether we still need to honour user supplied pauser

tgd (changes requested):
Just have a look at the pauser

peter-lawrey-admin (approved):

PR #960, opened 2021-11-29
Fix for unbalanced onAcquired()/onReleased() calls

garikjan:
As soon as a store is acquired in try-with-resources, that store is released using direct call to SingleChronicleQueueStore::close(). And onRelease() is not executed.

Issue #957, opened 2021-11-29
Update readme file so that JIRA link (at the end) is removed


minborg:
![image](https://user-images.githubusercontent.com/7457876/143839144-2eedc02b-5d1d-4282-84e6-d5803461b734.png)

JerryShea:
Done with https://github.com/OpenHFT/Chronicle-Queue/commit/788d24c0ae86452a8994a88d60147a03329f9d37

Issue #956, opened 2021-11-28
Setting the rollTime doesn't change the epoch.


peter-lawrey:
net.openhft.chronicle.queue.impl.single.SingleChronicleQueueBuilderTest#testReadMarshallable assumes the rollCycle sets the epoch.

alamar:
Looks like it's not needed in the light of https://github.com/OpenHFT/Chronicle-Wire/pull/362

alamar:
Issue no longer manifests.

PR #928, opened 2021-10-16
Test that all our RollCycle patterns fulfil the chronological == lexicographic ordering contract

nicktindall:

PR #1716, opened 2026-07-21
QUEUE-143: Source branch for CQE roll cleanup migration [do not merge]

peter-lawrey:
Source/provenance only; do not merge this as Queue cleanup APIs or CLI delivery. The retained source head is `f72b6737050d1ab6a124a92c3f07ab94d0e93e1c`, based on refreshed #1724 `dd06188703757dbac0172569e0b821fb967f02f2`. Both include current develop `4707043f3b1f9a1676a0514ccd3e65afe7ac44ce`.

Transfer ownership:

- Modern policy-free Queue integration belongs to #1745, followed by #1746 listeners; #1724/#1725 remain references.
- Archive primitives and corrected exclusion/generation/current-roll safeguards belong to CQE #937-#944 and #960. #960 must be refreshed and requalified against the final #956+#957 completion composition; its stale receipts do not qualify that composition.
- Named-tailer retention floor/parking eligibility and final physical/EOF generation-aware policy remain residual work for the frozen CQE #928 rewrite. They are not certified transferred by #960.

The local restructuring pack preserves a Git bundle, exact diff, content hashes and a scenario inventory for all four source test classes, including the refreshed invalid-name CLI regression. Unmatched scenarios remain explicitly assigned to the residual rewrite, not silently retired. No source ref is deleted and no new OSS cleanup child is created.

The prior refresh's full verification qualifies the preserved source tree only. Source closure requires the outstanding scenario-to-destination/reason ledger and final CQE qualification; that work is not implied complete by this review restructuring.


src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java
```diff
@@ -1265,6 +1280,96 @@ public TableStore<SCQMeta> metaStore() {
...
+            return false;
+        if (isReservedNamedTailerId(name))
+            throw reservedNamedTailerIdException(name);
+        if (name.startsWith(REPLICATED_NAMED_TAILER_PREFIX))
+            return false;
+        try (final ScopedResource<Bytes<Void>> bytesTl = acquireBytesScoped()) {
+            Bytes<Void> bytes = bytesTl.get().clear().append("index.").append(name);
+            LongValue longValue = tableStoreAcquireOrGet(bytes, 0, false);
+            if (longValue == null) return false;
+            longValue.setOrderedValue(0);
```
james-mcsherry, line 1414: If someone parks a tailer that turns out to be alive, would they get `true` back with no signal that the park was undone a moment later? Is it worth the javadoc saying the consumer must be stopped first?

PR #1720, opened 2026-08-07
SUPERSEDED: QUEUE-144 Queue context listener core

peter-lawrey:
> [!IMPORTANT]
> **Superseded: do not merge this PR.**
>
> The current replacement is [#1725](https://github.com/OpenHFT/Chronicle-Queue/pull/1725), based on the minimal QUEUE-143 Queue integration branch from [#1724](https://github.com/OpenHFT/Chronicle-Queue/pull/1724).

## Why this PR cannot simply be retargeted

This branch is descended from the obsolete appender prerequisite [#1719](https://github.com/OpenHFT/Chronicle-Queue/pull/1719), which was itself based on the old full `feature/QUEUE-143-RollFileCleanupMain` branch. Retargeting this head to the new QUEUE-143 integration branch would make removed cleanup implementation appear in the merge diff.

PR #1725 reapplies the context-listener core patch and merges `feature/QUEUE-143-queue-integration-points`. Its configured base is an ancestor of its head, and its merge diff is limited to the intended eleven Queue context-listener core files.

## Review findings carried to the replacement stack

- The core patch still says the supplied method writer is callback-scoped and implements that restriction. This conflicts with the final Wire contract, under which the writer may be retained for normal use.
- Listener failure before writing currently retries in the same roll context, while the Wire contract says a failed notification is not retried for that context.
- This core layer adds roughly 1,000 lines of production behaviour but its only test class injects a recording lifecycle. The black-box roll, persistence, failure and ownership tests are deferred to [#1726](https://github.com/OpenHFT/Chronicle-Queue/pull/1726). At least a minimal successful-write/once-per-roll test should accompany the core change, or #1725 and #1726 must be reviewed and merged as one behavioural unit.

Review and merge should continue on #1725 after those contract differences are reconciled.

PR #1721, opened 2026-08-07
SUPERSEDED: QUEUE-144 Queue context listener behaviour tests

peter-lawrey:
> [!IMPORTANT]
> **Superseded: do not merge this PR.**
>
> This PR is stacked on the obsolete core branch from closed [#1720](https://github.com/OpenHFT/Chronicle-Queue/pull/1720). The current replacement is [#1726](https://github.com/OpenHFT/Chronicle-Queue/pull/1726), whose base is the correctly split core PR [#1725](https://github.com/OpenHFT/Chronicle-Queue/pull/1725).

## Why this PR cannot simply be retargeted

The head retains the old QUEUE-143/full-cleanup ancestry through #1720 and #1719. Retargeting it directly to #1725 would expose removed cleanup changes in the merge diff. PR #1726 recreates this fixture layer on the correct split base. Its configured base is an ancestor of its head and its diff is limited to the intended three test files.

## Correctness findings carried to PR #1726

- `listenerFailureBeforeWritingRetriesOnNextWrite` and `appenderUsableAfterListenerError` require a retry in the same roll context. The finalized Wire contract is at-most-once: a failed notification is not retried until the context count increases.
- `suppliedWriterCannotBeUsedAfterCallbackReturns` asserts the superseded callback-scoped writer contract. The finalized contract allows the supplied writer to be retained and used normally.
- The double-buffer tests treat any `RuntimeException` from `DocumentContext.index()` as proof of buffering. They should require `IndexNotAvailableException`; otherwise unrelated failures satisfy the test.
- `stressTwoAppendersOneContextRecordPerRoll` checks only the total number of context records. One roll with two records and another with none would still pass. Group assertions by cycle and require exactly one context record in every cycle.
- The corrupt-index handling branch is explicitly left untested as “fragile”. That is a real failure path and needs a deterministic injected or mocked test rather than inspection-only coverage.

The exact persisted-output fixtures and representative serialised DTOs are useful, but these assertions must be aligned with the final contract before the replacement test PR is accepted.

### DIFF
--- a/src/test/java/net/openhft/chronicle/queue/ContextListenerBugTest.java
+++ b/src/test/java/net/openhft/chronicle/queue/ContextListenerBugTest.java
@@ -1,0 +1,421 @@
+/*
+ * Copyright 2013-2025 chronicle.software; SPDX-License-Identifier: Apache-2.0
+ */
+package net.openhft.chronicle.queue;
+
+import net.openhft.chronicle.core.io.Closeable;
+import net.openhft.chronicle.core.time.SetTimeProvider;
+import net.openhft.chronicle.core.time.SystemTimeProvider;
+import net.openhft.chronicle.queue.impl.single.SingleChronicleQueueBuilder;
+import net.openhft.chronicle.wire.DocumentContext;
+import net.openhft.chronicle.wire.MarshallableOut;
+import net.openhft.chronicle.wire.Wire;
+import net.openhft.chronicle.wire.WireType;
+import org.junit.After;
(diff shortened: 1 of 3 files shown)


peter-lawrey:
Superseded by the split-base QUEUE-144 stack. Use OpenHFT/Chronicle-Queue#1726 instead; it is stacked on #1725/#1724 and does not include the moved QUEUE-143 roll cleanup implementation.

PR #1734, opened 2026-08-20
[#1096] Inconsistent EOF handling 

### LINKED ISSUE #1096: Inconsistent EOF handling
[SingleChronicleQueueStore](https://github.com/OpenHFT/Chronicle-Queue/blob/ea/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueueStore.java) handles writing an EOF when a cycle rolls.

The implementation has a `tryReserve` check, and either uses the already mapped wire (if reserve ok), or remaps a temporary wire (if reserve not ok).

The temporary wire differs in at least 2 important respects vs the already-mapped wire, which leads to inconsistent EOF behaviour. Specifically:
- usePadding is true for the local mapping, false for the temp mapping
- the writePosition() is off by 8 bytes between the local (higher value) and temp mapping

````
    public boolean writeEOF(@NotNull Wire wire, long timeoutMS) {
        throwExceptionIfClosed();

        String fileName = mappedFile.file().getAbsolutePath();

        // just in case we are about to release this
        if (wire.bytes().tryReserve(this)) {
            try {
                return writeEOFAndShrink(wire, timeoutMS);

            } finally {
                wire.bytes().release(this);
            }
        }

        try (MappedBytes bytes = MappedBytes.mappedBytes(mappedFile.file(), mappedFile.chunkSize())) {
            Wire wire0 = WireType.valueOf(wire).apply(bytes);
            return writeEOFAndShrink(wire0, timeoutMS);

        } catch (Exception e) {
            Jvm.warn().on(getClass(), "unable to write the EOF file=" + fileName, e);
            return false;
        }
    }
````

peter-lawrey:
## Refresh on 12 September 2026

Merged develop `4707043f3b1f9a1676a0514ccd3e65afe7ac44ce` into the existing branch with normal forward commits. Updated head: `1dfd0dd27569837510c08b0bc8eab5c6ba3bcf7f`. Both develop `4707043f3b1f9a1676a0514ccd3e65afe7ac44ce` and the declared parent are ancestors of this head; both missing-commit counts are zero.

Retained current EOF/recovery changes and the regression assertions for actual positions, padding and next-header bytes. Full clean verify passed on Java 21: 1,115 reported tests, 52 skipped.

Supported-platform CI and composition with the intended Queue/Wire delivery remain required.

Local runs used Linux x86-64 and OpenJDK 21.0.12 unless Java 8 is explicitly stated (8u502). Reported test counts include skips. Commands requested zero Surefire reruns. Draft status and declared target are retained.

Earlier implementation/review notes follow. Earlier validation and performance claims apply only to their stated revisions and are superseded by the current receipt above where they differ.

---

## What changed

Local `fad5a2d0d` makes fallback EOF writing use the normal padding/position path with forced-reservation-failure coverage.

## Why

This publishes the reviewed local solution for [#1096](https://github.com/OpenHFT/Chronicle-Queue/issues/1096) so the implementation can be discussed in the normal review workflow.

## Review status

**Draft — coherent local solution, not yet merge-ready.**

Before marking ready: Rerun byte/position comparison and publish.

## Validation

- `git diff --check` against `ea`: passed.
- Internal `.pr/` scaffolding and local workspace paths: removed.
- Focused validation recorded by the backlog audit is described above; it was not rerun during this publication pass.
- Remote CI and maintainer review are still required.

## Tracking

Fixes #1096


peter-lawrey:
Writing to unpadded files is unsafe and not supported

PR #1739, opened 2026-08-24
QUEUE-146: keep ordinary appenders on the latest published cycle

peter-lawrey:
Keep the authoritative-cycle invariant in one delivery PR. Use five ordered review chapters, preserving the descendant chain #1740 -> #1741 -> #1745 -> #1746:

1. `RollingResourcesCache` weekly filename decoding, locale and year-boundary round trips. Alternate epochs require their own controls.
2. Transactional binding acquisition and ownership rollback.
3. Document-acquisition `Error` cleanup: nesting restoration and unlocking stay together.
4. Private-buffer rollback/failed-flush cleanup: lock ownership and failure-path clearing stay together.
5. Authoritative cycle publication and selection: UInt31/sentinel compatibility, persisted decoding, minimum-before-maximum publication, logical directory bounds, missing-published-roll preflight and forward-only selection, with all consumers and the regression matrix.

The selector's temporary-store inspection/event pairing belongs to chapter 5. It is not automatically an independently reproducible base ownership fix. A new published prerequisite stack is deferred unless an identified backport consumer justifies requalifying all descendants.

Head `5ee1a25d320480de42841a39b75f8d68461c12ee` and its staged base are unchanged by this review-map update.

<details>
<summary>Existing implementation and validation receipt (restructuring changes only the review map)</summary>

## Purpose

Q1 of QUEUE-146: keep ordinary appenders at or above the highest cycle published by any writer, including when wall time moves backwards, without adding directory scans to the steady-state append path.

## Contract

- Ordinary writes select the greater of the wall-clock cycle and the latest published cycle.
- Internal cycle state is either a UInt31 value (`0` through `Integer.MAX_VALUE`) or Wire's canonical `MarshallableOut.UNSET_CONTEXT`; both cycle zero and `Integer.MAX_VALUE` are valid.
- New writable initialisation and empty refreshes persist legacy `Integer.MIN_VALUE` so older readers still recognise an empty Queue. Persisted `long` values are decoded and range-checked before narrowing: raw `Long.MIN_VALUE`, legacy `Integer.MIN_VALUE` and `UNSET_CONTEXT` decode as unset; other values below `-1` or above `Integer.MAX_VALUE` fail closed.
- The legacy stored minimum may still be `Integer.MAX_VALUE`, so emptiness is determined from the decoded maximum rather than from the minimum alone.
[...]

Issue #1280, opened 2023-01-11
Remove automatic roll shrinking code


alamar:
It is not sufficiently tested and, worse of all, Tailer will un-shrking a roll when opening it.

Let's keep it as an utility which may be called for archiving purposes for now.

alamar:
See #1275

alamar:
See also #1276

alamar:
I'm not sure how to implement a stand-alone utility - currently, the implementation relies on knowing the end of stream position, later on we don't know such position and don't want to write extra code to iterate the queue and find it out.

Will submit complete removal of that code - may be reinstated once #1276 is taken care of in the future

hft-team-city:
Released in [Chronicle-Queue-5.24ea9](https://github.com/OpenHFT/Chronicle-Queue/releases/chronicle-queue-5.24ea9), [BOM-2.24ea43](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.24ea43)

PR #1740, opened 2026-08-24
QUEUE-146: roll ordinary appends past sealed cycles

peter-lawrey:
## Purpose

Q2 of QUEUE-146: retain EOF as a hard seal while allowing an ordinary append to advance exactly once from Q1's authoritative destination.

## Contract

- Document and sequential-byte ordinary writes share the same bounded EOF transition.
- Q1 selects the published destination first; EOF then advances from that destination to `cycle + 1`, without creating intermediate cycles.
- A second EOF propagates. `Integer.MAX_VALUE` is rejected before successor creation or persistent source/listing mutation.
- At this PR's own head, exact-index writes remain strict and never invoke ordinary successor retry. Q3 (#1741) separately introduces validated exact recovery.
- Queue clears transient Wire header-acquisition state for compatible Wire versions that predate Wire-side EOF cleanup. Current staged Wire self-cleans, so tests using it do not discriminate this compatibility fallback.

Listener-specific preflight remains in #1746.

- `stalledWriterAdvancesOnceFromPublishedSealedCycle` combines Q1 and Q2: parked cycle 0, another writer's published/sealed cycle 3, then cycle 4; no cycles 1 or 2.
- `eofAdvanceRejectsCycleOverflowBeforeMutation` checks publication, listing modification count, filenames, source write position and EOF word.
- The Windows cleanup test no longer catches `Throwable` or silently accepts failed deletion. Only deletion is retried, for at most five seconds, using the existing strict deletion helper. Persistent failure and unrelated errors fail the test. Linux deletion is also strict. This test-only correction needs no artificial production `//!`.

## Current ancestry and validation

- Base: `5ee1a25d320480de42841a39b75f8d68461c12ee` (`feature/QUEUE-143-monotonic-ordinary-append-cycle`).
- Head: `9bd9b7e6db80b7d037a4f4a0f52f976c7dd4a4fe`.
- Base equals merge base; 3 changed files and no POM delta.
- Existing branch history is preserved with a merge commit and a normal fast-forward push. The tree matches tested candidate `d0c0ff48cb9654dd547c760cb1c88d20ca4789f6`.
- The net change is again Q2's intended three files. Its patch ID matches the original Q2-only comparison. Old commits remain in history; the refreshed #1739 is an ancestor and stale Q1 changes are absent from the net diff.
- Full verification with the source-built Wire dependency: **1,164 tests, zero failures/errors, 52 skipped**. The inherited producer-join-only test correction also passed `StoreTailerTest` (25 tests) and Maven verification checks.
- Linux amd64/OpenJDK 21.0.12, Maven 3.9.11, offline isolated Maven repository, `-Dsurefire.rerunFailingTestsCount=0`. Required build checks passed.
- Wire runtime `2026.10-SNAPSHOT` was rebuilt from merged #1280 commit `3dd7fe5983c0133703ee72f57bced9a25270eb6f`; JAR SHA-256 `92803fb011f5071900f5c3b1b637cf39f6254e86b4e29f77f1256ecc96dde755`. This is source-build evidence for the local run; released immutable coordinates remain a gate.
- No Windows, macOS or ARM execution is claimed.

## Stack and review gates

The repaired order is `staged/develop` -> #1739 -> #1740 -> #1741. #1741 remains draft for final-stack review. The coordinate work in #1750 and follow-ups #1745/#1746 are unchanged.

Wire #1280 merged on 2026-09-07. Successful CI on the published heads, independent review and immutable release/BOM coordinates remain outstanding. Previous-head results do not qualify the new heads.

The shared rolling-test worker lifetime and failure-reporting repair is owned by [Queue #1756](https://github.com/OpenHFT/Chronicle-Queue/pull/1756). Incorporate and revalidate that repair through the declared prerequisite order. The underlying write-lock stall and this PR’s normal merge requirements remain separate.


src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
```diff
@@ -701,7 +701,38 @@ private void moveToCycleForAppend() {
      * @param safeLength the safe length of data that can be written
      * @return the position of the written header
      */
-    private long writeHeader(@NotNull final Wire wire, final long safeLength) {
+    private long writeHeaderForOrdinaryAppend(final long safeLength) {
+        assert wire != null;
+        for (; ; ) {
```
tgd: Suggest adding bound to this to prevent infinite loop in case semantics inside writeHeader change in the future

tgd (approved):
Approved with suggestion

src/main/java/net/openhft/chronicle/queue/impl/single/StoreAppender.java
peter-lawrey: no longer uses a loop

peter-lawrey:
Restack/review update for exact head `67db828a08b6b1652f345cae36683de671d9b2a9`, a literal descendant of #1739 `f036832bc`.

This keeps ordinary destination selection and the single EOF advance in QUEUE-146. The ordinary and QUEUE-144 listener paths now consume the same destination preflight, and cycle overflow is rejected before advancing.

No QUEUE-143 maintenance policy or QUEUE-144 listener lifecycle was moved into this PR. Complete-directory deletion/reset is also intentionally out of scope; historical maintenance preserves the latest roll and the persistent write floor.

The focused composed Queue regression suite passes locally. TeamCity is still running, so this is not a merge-readiness claim.

